PR #2, opened 2017-03-23
Hero

Blackbaud-BrandonJones:
screen shot 2017-03-23 at 2 16 38 pm


src/modules/hero/hero.component.html
```diff
@@ -1,8 +1,9 @@
 <div class="stache-hero" 
-  [ngStyle]="{ 'background-image': 'url(' + backgroundImageUrl + ')' }">
+  [ngStyle]="{'background-image': 'url(' + backgroundImageUrl + ')' }">
```
Blackbaud-SteveBrush: Spaces inside curly braces `{ o }`.

src/modules/hero/hero.component.scss
```diff
@@ -11,21 +10,20 @@
...
 .stache-hero-overlay {
   background-color: #000;
-  opacity: .35;
+  opacity: 0.4;
   position: absolute;
-  top: 0;
-  right: 0;
-  bottom: 0;
-  left: 0;
+  height: 100%;
```
Blackbaud-SteveBrush: You can add height and width, here, but the positional properties (top, left, etc.) are also important for cross-browser support.

src/modules/hero/hero.component.spec.ts
```diff
@@ -0,0 +1,80 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { DebugElement } from '@angular/core';
```
Blackbaud-SteveBrush: Separate libraries with new lines so that they have logical groupings (angular, blackbaud, internal).

src/modules/hero/hero.component.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    component.overlayOpacity = '0.5';
+    expect(component.overlayOpacity).toEqual('0.5');
+    component.overlayOpacity = '.03';
+    expect(component.overlayOpacity).toEqual('0.03');
+    component.overlayOpacity = '0.12456';
+    expect(component.overlayOpacity).toEqual('0.12');
+  });
+
+  it('should default to 0.4 given an invalid input', () => {
+    component.overlayOpacity = 'Sam Wise Gamgee';
```
Blackbaud-SteveBrush: Because we're writing a public-facing library, it's a better practice to create "boring" tests, and avoid humor/cultural references. In other words, use something like `invalid value` instead of `Sam Wise Gamgee` :)

I think Hare and myself wrote a few of these in a similar fashion before; we'll need to go back and clean those up before the demo, as well. I'll create an issue for that.

src/modules/hero/hero.component.ts
```diff
@@ -15,5 +16,43 @@ export class StacheHeroComponent {
...
+  private _parseOpacityValueToDecimal(opacityValue: string): string {
+    let percentageCandiate = parseInt(opacityValue, 10);
+    if (opacityValue.includes('.')) {
+      percentageCandiate = Math.round(parseFloat(opacityValue) * 100) / 100;
+    }
+    let validDecimal = this._validateOpacityDecimal(percentageCandiate);
+    return String(validDecimal);
+  }
+
+  private _validateOpacityDecimal(value: number): number {
```
Blackbaud-SteveBrush: Angular [recommends](https://angular.io/styleguide#!#03-04) NO underscores for private properties and methods. In the case of a getter/setter, an underscore is valid.

src/modules/hero/hero.component.ts
```diff
@@ -15,5 +16,43 @@ export class StacheHeroComponent {
...
+    return this._overlayOpacity;
+  }
+
   private _backgroundImageUrl = 'none';
+  private _overlayOpacity = '0.6';
+
+  private _parseOpacityValueToDecimal(opacityValue: string): string {
+    let percentageCandiate = parseInt(opacityValue, 10);
+    if (opacityValue.includes('.')) {
+      percentageCandiate = Math.round(parseFloat(opacityValue) * 100) / 100;
```
Blackbaud-SteveBrush: percentageCandi _d_ ate?

src/modules/hero/hero.component.ts
```diff
@@ -15,5 +16,43 @@ export class StacheHeroComponent {
...
+    let validDecimal = value;
+    if (value > 100) {
+      validDecimal = 1;
+    } else if (value < 0) {
+      validDecimal = 0;
+    } else if (isNaN(value)) {
+      validDecimal = 0.40;
+    }
+
+    if (validDecimal > 1 ) {
```
Blackbaud-SteveBrush: Remove space after `1`

src/modules/nav/nav.component.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { Router } from '@angular/router';
 
-import { expect } from '@blackbaud/skyux-builder/runtime/testing/browser';
+import { expect } from '@blackbaud/skyux-builder/runtime/testing/browser/';
```
Blackbaud-SteveBrush: Probably don't need that slash anymore, eh?

src/modules/hero/hero.component.ts
```diff
@@ -6,6 +6,7 @@ import { Component, Input } from '@angular/core';
   styleUrls: ['./hero.component.scss']
 })
 export class StacheHeroComponent {
+
   @Input()
   set backgroundImageUrl(imageUrl: string) {
```
Blackbaud-SteveBrush: Getter/setter for `backgroundImageUrl` not really needed, here.

src/modules/hero/hero.component.ts
```diff
@@ -15,5 +16,43 @@ export class StacheHeroComponent {
...
+  }
+
+  private validateOpacityDecimal(value: number): number {
+    let validDecimal = value;
+    if (value > 100) {
+      validDecimal = 1;
+    } else if (value < 0) {
+      validDecimal = 0;
+    } else if (isNaN(value)) {
+      validDecimal = 0.40;
```
Blackbaud-SteveBrush: This default value is different from the `0.6` set above. Maybe create a `readonly` property for the default?

src/modules/hero/hero.component.ts
```diff
@@ -15,5 +16,43 @@ export class StacheHeroComponent {
...
+
+  get overlayOpacity(): string {
+    return this._overlayOpacity;
+  }
+
   private _backgroundImageUrl = 'none';
+  private _overlayOpacity = '0.6';
+
+  private parseOpacityValueToDecimal(opacityValue: string): string {
+    let percentageCandiate = parseInt(opacityValue, 10);
```
Blackbaud-SteveBrush: "Candiate"

src/modules/hero/hero.component.ts
```diff
@@ -15,5 +16,43 @@ export class StacheHeroComponent {
...
+    if (opacityValue.includes('.')) {
+      percentageCandiate = Math.round(parseFloat(opacityValue) * 100) / 100;
+    }
+    let validDecimal = this.validateOpacityDecimal(percentageCandiate);
+    return String(validDecimal);
+  }
+
+  private validateOpacityDecimal(value: number): number {
```
Blackbaud-SteveBrush: I was initially confused by the naming of this method and had to read through it to figure out what it was doing. It's really just parsing a number into an [interval](https://en.wikipedia.org/wiki/Interval_(mathematics)). What about...

```
// Converts a number to a value between 0 and 1.
private parseInterval(value: number): number {
  let interval = parseFloat(value);

  if (isNaN(interval)) {
    return 0.4;
  }

  if (interval > 100) {
    return 1;
  }

  if (interval < 0) {
    return 0;
  }

  if (interval % 1 > 0) {
    return interval;
  }

  return interval / 100;
}
```

src/modules/hero/hero.component.ts
```diff
@@ -7,13 +7,41 @@ import { Component, Input } from '@angular/core';
...
+  get overlayOpacity(): string {
+    return this._overlayOpacity || this.defaultOpacity;
   }
 
-  private _backgroundImageUrl = 'none';
+  private readonly defaultOpacity: string = '0.4';
+  private _overlayOpacity: string;
+
+  private parseInterval(value: string): number {
+    let interval = Math.round(parseFloat(value) * 100) / 100;
```
Blackbaud-SteveBrush: Curious why you're using `Math.round`? Isn't `0.0435322` a valid opacity value?

src/modules/hero/hero.component.ts
```diff
@@ -7,13 +7,41 @@ import { Component, Input } from '@angular/core';
 })
 export class StacheHeroComponent {
   @Input()
-  set backgroundImageUrl(imageUrl: string) {
-    this._backgroundImageUrl = imageUrl;
+  public backgroundImageUrl: string;
+
+  @Input()
+  set overlayOpacity(opacityCandidate: string) {
```
Blackbaud-SteveBrush: This method is a lot better than it was, but I wouldn't be afraid to simplify it further.

Also, `toString` makes more immediate sense than `String()`...

```
@Input()
set overlayOpacity(value: string) {
  let sanitized = value.replace(/[^\d.-]/g, '');
  let interval = this.parseInterval(sanitized);
  this._overlayOpacity = interval.toString();
}
```

PR #4, opened 2017-03-24
Forward And Back Buttons

Blackbaud-BrandonJones:
screen shot 2017-03-24 at 10 55 41 am


Blackbaud-SteveBrush:
This is a subjective complaint, but I'm struggling a little bit with the name of this component. "Navigation buttons" could be confused with a collection of any number of buttons, which link to other pages, instead of what it's intended to be: a pair of buttons that traverse a sibling navigation tree.

The first UI pattern that came to mind when I looked at your code was "pagination". Pagination traditionally has a Previous and a Next button, as well as a list of numbers representing each page in the collection. We _could_ call this the Pagination Component and have configuration to show/hide the numbers in between the previous/next buttons (the previous/next buttons still showing the page titles for their labels).

Just some thoughts for the morning!

src/styles/_variables.scss
```diff
@@ -1,4 +1,5 @@
 @import '_sky-variables.scss';
+@import '_font-awesome-variables';
```
Blackbaud-SteveBrush: It would be better to pull the variables directly from the node_modules folder. (I added the SKY UX variables to our source because there's no way to get the variables dynamically, as of yet.)

src/modules/nav/nav.module.ts
```diff
@@ -2,6 +2,8 @@ import { NgModule } from '@angular/core';
 import { CommonModule } from '@angular/common';
 import { RouterModule } from '@angular/router';
 
+import { StacheNavComponentService } from './nav.component.service';
```
Blackbaud-SteveBrush: I'm not sold on a nav component service. Brainstorm an idea that doesn't require it. Maybe a component `@Output()`?

src/styles/_variables.scss
Blackbaud-BrandonJones: so we can do that, though their variables include an _nav_   in the variable names for each icon.    eg.   $fa_var_cogs    $fa-var-angle-double-right

src/modules/forward-back-buttons/forward-back-buttons.component.scss
```diff
@@ -0,0 +1,32 @@
...
+
+    .stache-nav-list {
+      justify-content: space-between;
+      margin: 10px;
+      flex-direction: row;
+
+      li:first-child {
+        span {
+          &:before {
+            @include import_font_awesome_icon($fa-angle-double-left);
```
Blackbaud-SteveBrush: Curious. Where is this mixin coming from?

Blackbaud-BrandonJones: This is coming from the mixin I made in _font-awesome-variables.scss,  should it be moved into a separate _stache-mixins.scss file?

src/app/components/demo-component.service.ts
```diff
@@ -191,6 +191,25 @@ export class StacheDemoComponentService {
...
     },
     {
+      name: 'Forward Back Buttons',
+      route: '/components/forward-back-buttons',
+      summary: '',
+      getCodeFiles: () => {
+        return [
+          {
+            name: 'forward-back-buttons-demo.component.html',
+            fileContents: require('!!raw!./forward-back-buttons/forward-back-buttons-demo.component.html'),
```
Blackbaud-SteveBrush: Might want to go ahead and add `-loader` to all your requires...

blackbaud-brandonhare: HAHAHAHAHA

Blackbaud-SteveBrush:
I'm beginning to think that maybe the forward and back buttons are doing too much. Their purpose is to display two routes as buttons; that's it. If we want a more "dynamic" approach to which routes should be represented, the parent component should be in charge of that. In our case, the parent component could be a "tutorial" layout, which owns the logic for the active page, and what the user has accomplished on their tutorial todo list.

So, really, the `StacheForwardAndBackButtonsComponent` should only accept an array of routes and print the first one and the last one, if they exist. [A component should try its best to only do one thing.](https://angular.io/styleguide#!#single-responsibility)

The extra functionality you have here--to select routes "dynamically"--could probably be migrated to the future Tutorial Layout.

src/app/learn/faq/index.html
```diff
@@ -0,0 +1,19 @@
...
+  [breadcrumbsRoutes]="[{ name: 'Home', path: '/' }, { name: 'Learn', path: '/learn' }]">
+
+  <stache-page-summary>
+    <p>
+      The SKY UX framework provides a consistent, cohesive user experience for Blackbaud products and for third-party developers. SKY UX provides the components, services, and modules that developers need to incorporate Blackbaud styles and patterns.
+    </p>
+  </stache-page-summary>
+
+  <stache-forward-back-buttons
+    [routes]="[{ name: 'Resources', path: '/learn/resources' }, undefined]">
```
Blackbaud-SteveBrush: I was under the impression that we were going to be explicit with the inputs, `previousRoute` and `nextRoute`?

Blackbaud-BrandonJones: We can do both,   as it gets set to an array as it is, I didn't see a reason not to allow the versatility of options?   Should be be more prescriptive here instead and only allow those two inputs?

src/modules/forward-back-buttons/forward-back-buttons.component.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { StacheNavModule } from '../nav/nav.module';
+import { StacheForwardBackButtonsComponent } from './forward-back-buttons.component';
+
+describe('Forward, Back Button Component', () => {
+  let component: StacheForwardBackButtonsComponent;
+  let fixture: ComponentFixture<StacheForwardBackButtonsComponent>;
+  let debugElement: DebugElement;
+  let mockRouter: any;
+      class MockRouter {
```
Blackbaud-SteveBrush: Extra tabs...

src/modules/forward-back-buttons/forward-back-buttons.component.ts
```diff
@@ -0,0 +1,22 @@
...
+import { StacheNavLink } from '../nav/nav-link';
+
+@Component({
+  selector: 'stache-forward-back-buttons',
+  templateUrl: './forward-back-buttons.component.html',
+  styleUrls: ['./forward-back-buttons.component.scss']
+})
+export class StacheForwardBackButtonsComponent implements OnInit, StacheNav {
+  @Input()
+    public previousRoute: StacheNavLink;
```
Blackbaud-SteveBrush: Extra tab...

Issue #5, opened 2017-03-24
Implement smooth scroll


Blackbaud-SteveBrush:
https://github.com/angular/angular/issues/6595

```
public scrollTo(element: any, speed: number = 60) {
  let to = element.offsetTop;
  let yPos = window.scrollY;
  const scroll = () => {
    setTimeout(() => {
      yPos += speed;
      window.scrollTo(0, yPos);
      if (yPos < to) {
        scroll();
      }
    }, 10);
  };
  scroll();
}
```

PR #19, opened 2017-03-29
Back-to-top & Smooth Scroll

blackbaud-brandonhare:
Implements #5 #23


src/modules/back-to-top/back-to-top.component.html
```diff
@@ -0,0 +1,3 @@
+<div #backToTop class="back-to-top" (click)="goToTop()" [ngStyle]="{'background-color': backgroundColor, 'color': color}">
```
Blackbaud-SteveBrush: The word `scroll` better represents the action taken here ("go" could imply a URL change).

Blackbaud-SteveBrush: Also, I don't think having an override for background color and text color is needed, here. We should be prescriptive on our styles. What were your thoughts on this feature?

Blackbaud-SteveBrush: Also, I don't think having an override for background color and text color is needed, here. We should be prescriptive on our styles. What were your thoughts on this feature?

blackbaud-brandonhare: Works for me. I'll change.

src/modules/back-to-top/back-to-top.component.scss
```diff
@@ -0,0 +1,31 @@
+.back-to-top {
+    position: fixed;
```
Blackbaud-SteveBrush: Only 2-spaces tabs for SCSS, please!

src/modules/back-to-top/back-to-top.component.scss
```diff
@@ -0,0 +1,31 @@
...
+    bottom: 10px;
+    right: 10px;
+    color: #FFF;
+    background: #1c84c6;
```
Blackbaud-SteveBrush: This color can be pulled as a variable from the `_variables.scss` file.

src/modules/back-to-top/back-to-top.component.html
blackbaud-brandonhare: My rationale for this is to allow the component to be used elsewhere. If a site has a specific design style, you wouldn't want the default color to clash.

src/modules/back-to-top/back-to-top.component.scss
```diff
@@ -0,0 +1,31 @@
...
+    text-align: center;
+    font-size: 2em;
+    line-height: 1;
+    opacity: 0;
+    transition: opacity .5s;
+}
+
+.back-to-top:hover,
+.back-to-top:focus {
+    background: #1a7bb9;
```
Blackbaud-SteveBrush: Same here, RE: color and variables.

src/modules/back-to-top/back-to-top.component.scss
```diff
@@ -0,0 +1,31 @@
...
+    cursor: pointer;
+}
+
+i.fa {
+    font-size: medium;
```
Blackbaud-SteveBrush: Can we be more explicit with the font size?

src/modules/back-to-top/back-to-top.component.scss
```diff
@@ -0,0 +1,31 @@
...
+    display: block;
+    margin-top: 5px;
+}
+
+.show {
```
Blackbaud-SteveBrush: Common classnames for this behavior are `.hidden` or `.visible` (I prefer `.hidden` because an element is visible by default.)

src/modules/back-to-top/back-to-top.component.ts
```diff
@@ -0,0 +1,35 @@
+import { Component, ElementRef, Input, OnInit, ViewChild } from '@angular/core';
+
+@Component({
+  selector: 'stache-back-to-top',
+  templateUrl: './back-to-top.component.html',
+  styleUrls: ['./back-to-top.component.scss']
+})
+export class StacheBackToTopComponent {
+  @Input() offset: number;
```
Blackbaud-SteveBrush: Stack decorators on top of their property:
```
@Input()
public offset: number;
```

Blackbaud-SteveBrush: Be explicit with both property and method access (`public`, `private`, etc.).

Blackbaud-SteveBrush: (There is a tslint config option for this somewhere, and I know SKY UX uses it on Builder. Let me attempt to apply that to our code base as well.)

src/modules/back-to-top/back-to-top.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  @Input() color: string;
+  @Input() backgroundColor: string;
+  @ViewChild('backToTop') el: ElementRef;
```
Blackbaud-SteveBrush: `el` --> `elementRef`

src/modules/back-to-top/back-to-top.component.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  ngOnInit(): void {
+    window.addEventListener('scroll', () => {
+      if (window.pageYOffset > this.offset) {
+        this.el.nativeElement.classList.add('show');
```
Blackbaud-SteveBrush: `classList` supported on all major browsers?

blackbaud-brandonhare: http://caniuse.com/#search=classList

src/modules/back-to-top/back-to-top.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  styleUrls: ['./back-to-top.component.scss']
+})
+export class StacheBackToTopComponent {
+  @Input() offset: number;
+  @Input() color: string;
+  @Input() backgroundColor: string;
+  @ViewChild('backToTop') el: ElementRef;
+
+  ngOnInit(): void {
+    window.addEventListener('scroll', () => {
```
Blackbaud-SteveBrush: It's dangerous to apply event listeners directly to the window in our components because we could potentially add multiple listeners in a given session. 

Look into implementing `@HostListener()`:
https://juristr.com/blog/2016/09/ng2-event-registration-document/

src/modules/back-to-top/back-to-top.component.ts
```diff
@@ -0,0 +1,35 @@
...
+      if (window.pageYOffset > this.offset) {
+        this.el.nativeElement.classList.add('show');
+      } else {
+        this.el.nativeElement.classList.remove('show');
+      }
+    });
+  }
+
+  goToTop() {
+    try {
```
Blackbaud-SteveBrush: Possible to avoid `try`/`catch` blocks?

blackbaud-brandonhare: If you can think of a way to add a fallback without it I'm open to suggestions. This seemed like the simplest solution to ensure the back to top worked regardless.

src/app/learn/index.html
```diff
@@ -1,2 +1,3 @@
 <app-nav></app-nav>
-<app-learn></app-learn>
\ No newline at end of file
+<app-learn></app-learn>
+<stache-back-to-top [offset]=200></stache-back-to-top>
```
Blackbaud-SteveBrush: Implement the `@TypeConverter()` decorator so that consumers can type `offset="200"`.

(Also: put `200` in double quotes.)

src/modules/back-to-top/back-to-top.component.html
Blackbaud-SteveBrush: Who would be using this outside of SKY UX2 SPA?

blackbaud-brandonhare: It could still be in a spa, but maybe they want to use the blackbaud green (or if we are being open source minded, their company's color), I think we should give them the option right?

src/modules/back-to-top/back-to-top.component.ts
Blackbaud-SteveBrush: Partial support for IE11. We'll definitely need to test this on IE before we release. Okay for now.

blackbaud-brandonhare: The partial support is for multiple parameters and specific types of DOM elements, neither of which I am using with this implementation. It should be fine. But agreed, we should test.

src/modules/nav/nav.component.ts
```diff
@@ -43,6 +43,8 @@ export class StacheNavComponent implements OnInit, StacheNav {
 
     if (route.fragment) {
       extras.fragment = route.fragment;
+      let ref = document.getElementById(route.fragment);
```
Blackbaud-SteveBrush: `ref` --> `element`

src/modules/back-to-top/back-to-top.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  public color: string;
+
+  @Input()
+  public backgroundColor: string;
+
+  @ViewChild('backToTop')
+  public elementRef: ElementRef;
+
+  @HostListener('window:scroll')
+  onWindowScroll() {
```
Blackbaud-SteveBrush: `public`, please!

src/modules/page-anchor/page-anchor.component.ts
```diff
@@ -29,6 +29,7 @@ export class StachePageAnchorComponent implements OnInit, StacheNavLink {
   }
 
   public addHashToUrl(): void {
+    this.elementRef.nativeElement.scrollIntoView({behavior: 'smooth'});
```
Blackbaud-SteveBrush: Spaces between curly braces. `{ behavior: 'smooth' }` (This needs to be added to the tsling.config, too. <-- working on it)

src/modules/back-to-top/back-to-top.component.scss
```diff
@@ -0,0 +1,33 @@
+@import '../../styles/_variables';
+
+.back-to-top {
+  position: fixed;
+  bottom: 10px;
+  right: 10px;
+  color: $sky-color-white;
```
Blackbaud-SteveBrush: Sorry, this is my fault for not explaining correctly: we should always use "stache" variables in our "stache" stylesheets. These variables can certainly reference "sky" variables in the `_variables.scss` file, but it helps isolate features between the two libraries (see `$stache-color-blue` as an example). 

For example, what if SKY UX decided to change the variable `$sky-color-white` to `$skyux-white`? We'd have to do a find and replace in all of our stylesheets to make the change.

src/modules/back-to-top/back-to-top.component.ts
```diff
@@ -0,0 +1,43 @@
+import { Component, ElementRef, HostListener, Input, ViewChild } from '@angular/core';
+import { InputConverter } from '../shared';
```
Blackbaud-SteveBrush: Separate Angular imports from our proprietary imports with a new line.

src/modules/back-to-top/back-to-top.component.html
Blackbaud-SteveBrush: Stache Components rely on SKY UX styles, so in a sense they require SKY UX. I don't see a situation where Stache Components would be used outside of SKY UX SPA. Also, we're not providing color configuration anywhere else, so it seems a bit out of place. 

While having adaptable styles is forward-thinking in some respects, it adds complication and inconsequential configuration to something that should be relatively simple and prescriptive. If this feature is asked for in the future by multiple teams, it might be worth considering. Even still, if they wanted to style the component differently, we're exposing the classnames on the front end (`.stache-back-to-top`).

We should develop with extensibility in mind, but not to add features for the sake of having features.

src/modules/back-to-top/back-to-top.component.html
```diff
@@ -0,0 +1,3 @@
+<div #backToTop class="back-to-top hidden" (click)="scrollToTop()">
```
Blackbaud-SteveBrush: Stache's classnames should always be prefixed with `stache-` (`stache-back-to-top`).

src/modules/back-to-top/back-to-top.component.ts
```diff
@@ -0,0 +1,34 @@
...
+  @InputConverter()
+  public offset: number;
+
+  @ViewChild('backToTop')
+  public elementRef: ElementRef;
+
+  @HostListener('window:scroll')
+  public onWindowScroll() {
+    if (window.pageYOffset > this.offset) {
+      this.elementRef.nativeElement.classList.remove('hidden');
```
Blackbaud-SteveBrush: Alternative, instead of worrying about the `elementRef`:

Component HTML:
```
<button 
  class="stache-back-to-top"
  [ngClass]="{ 'hidden': isHidden }"
  title="Scroll to the top of the page.">
  <i class="fa fa-chevron-down"></i>
</button>
```

Component script:
```
@HostListener('window:scroll')
public onWindowScroll() {
  this.isHidden = (window.pageYOffset < this.offset);
}
```

src/modules/back-to-top/back-to-top.component.ts
```diff
@@ -0,0 +1,34 @@
...
+    } else {
+      this.elementRef.nativeElement.classList.add('hidden');
+    }
+  };
```
Blackbaud-SteveBrush: Remove semicolon.

src/modules/back-to-top/back-to-top.component.scss
```diff
@@ -0,0 +1,33 @@
...
+}
+
+i.fa {
+  font-size: 16px;
+  display: block;
+  margin-top: 5px;
+}
+
+.hidden {
+  opacity: 0;
```
Blackbaud-SteveBrush: The button is still clickable, even though its opacity is set to zero. Add `display: none`?

blackbaud-brandonhare: Display none doesn't animate.

src/modules/back-to-top/back-to-top.component.html
```diff
@@ -0,0 +1,7 @@
+<div #backToTop
```
Blackbaud-SteveBrush: `#backToTop` <-- no longer needed?

src/modules/back-to-top/back-to-top.component.html
```diff
@@ -0,0 +1,7 @@
...
+  class="stache-back-to-top"
+  [ngClass]="{ 'hidden': isHidden }"
+  (click)="scrollToTop()">
+
```
Blackbaud-SteveBrush: Mind if we remove this line-break?

src/modules/back-to-top/back-to-top.component.scss
```diff
@@ -0,0 +1,34 @@
...
+  height: 30px;
+  width: 30px;
+  text-align: center;
+  font-size: 2em;
+  line-height: 1;
+  opacity: 1;
+  transition: opacity .5s;
+}
+
+.stache-back-to-top:hover,
```
Blackbaud-SteveBrush: Throw this block into the `.stache-back-to-top` above:

```
.stache-back-to-top {
  &:hover,
  &:focus {}
}
```

src/modules/back-to-top/back-to-top.component.ts
```diff
@@ -0,0 +1,29 @@
...
+
+@Component({
+  selector: 'stache-back-to-top',
+  templateUrl: './back-to-top.component.html',
+  styleUrls: ['./back-to-top.component.scss']
+})
+export class StacheBackToTopComponent {
+  @Input()
+  @InputConverter()
+  public offset: number;
```
Blackbaud-SteveBrush: Let's put a default here so that folks won't need to supply one every time.

src/modules/nav/nav.component.ts
```diff
@@ -43,6 +43,8 @@ export class StacheNavComponent implements OnInit, StacheNav {
 
     if (route.fragment) {
       extras.fragment = route.fragment;
+      let element = document.getElementById(route.fragment);
```
Blackbaud-SteveBrush: Should probably check to make sure the element exists before attempting to use `scrollIntoView`, as it would throw an error.
```
if (element) { ... }
```

Blackbaud-SteveBrush: ^ Also something we should add to our test spec :)

src/app/learn/index.html
```diff
@@ -1,2 +1,3 @@
 <app-nav></app-nav>
-<app-learn></app-learn>
\ No newline at end of file
+<app-learn></app-learn>
+<stache-back-to-top offset="200"></stache-back-to-top>
```
Blackbaud-SteveBrush: The implementation and code are great! Next...

- [ ] Add this component to the sidebar and default layouts
- [ ] Create a variable on the wrapper, `showBackToTop` which defaults to `true`
- [ ] Write tests :)

PR #45, opened 2017-04-05
Moved mixins to separate folder

Blackbaud-SteveBrush:
- This is a common convention for mixins. Also, I removed them from the _variables file, and pulled them into the Forward Back styles explicitly.


Blackbaud-SteveBrush:
@Blackbaud-BrandonJones

Blackbaud-BrandonJones:
That works.  Is this better than having one Stache.scss  file entry that we import in any file that wants to use any mixin or variable?

Blackbaud-SteveBrush:
I think so, at least for now. It keeps our styles modular, and doesn't confuse as to what we're pulling into the file every time.

PR #46, opened 2017-04-05
fixed lint errors

Blackbaud-BrandonJones:


src/modules/wrapper/wrapper.component.ts
```diff
@@ -6,6 +6,7 @@ import { StachePageAnchorComponent } from '../page-anchor/page-anchor.component'
 import { StacheNavLink } from '../nav/nav-link';
 
 @Component({
+  // tslint:disable-next-line:component-selector-name
```
Blackbaud-SteveBrush: I'd prefer to place all linting configuration at the top of the file, to not clutter up the presentation.
```
/* tslint:disable:component-selector-name */
...
```

Blackbaud-BrandonJones: If we are good with disabling the rule for the entire file then that's fine.

src/modules/wrapper/wrapper.component.ts
```diff
@@ -1,3 +1,4 @@
+// tslint:disable:component-selector-name
```
Blackbaud-SteveBrush: Wrap it in "block" comments and then merge. `/* */`

PR #47, opened 2017-04-06
Added window provider.

blackbaud-brandonhare:
Created a window service and updated components that were using it to use the new service.


src/core.ts
```diff
@@ -34,6 +35,7 @@ smoothscroll.polyfill();
     StachePageHeaderModule,
     StacheLayoutModule,
     StacheWrapperModule
-  ]
+  ],
+  providers: [ WindowRef ]
```
Blackbaud-SteveBrush: Add the `WindowRef` provider to the individual modules that require it, instead of the root module (e.g., Back to Top).

src/modules/back-to-top/back-to-top.component.ts
```diff
@@ -14,13 +15,15 @@ export class StacheBackToTopComponent {
 
   public isHidden: boolean = true;
 
+  public constructor(private winRef: WindowRef) {}
```
Blackbaud-SteveBrush: `windowRef`, please.

src/core.ts
blackbaud-brandonhare: Hmm. The window reference is already used in 3 modules. I think having a global window reference makes sense.

Blackbaud-SteveBrush: Valid.

src/modules/shared/window.service.ts
```diff
@@ -0,0 +1,13 @@
+import { Injectable } from '@angular/core';
+
+function _window(): any {
```
Blackbaud-SteveBrush: I prefer the naming conventions used in the answer on the top of this page: http://stackoverflow.com/questions/34177221/angular2-how-to-inject-window-into-an-angular2-service

- Using "service" in the name of the service
- Not using the underscore for the method.
- Making `getWindow` a verb, instead of `_window()`

```
import { Injectable } from '@angular/core';

function getWindow(): any {
  return window;
}

@Injectable()
export class WindowRefService {
  get nativeWindow(): any {
    return getWindow();
  }
}
```

PR #49, opened 2017-04-10
Feature child sidebar nav

blackbaud-brandonhare:


src/app/learn/learn.component.ts
```diff
@@ -6,7 +6,10 @@ import { Component } from '@angular/core';
 })
 export class LearnComponent {
   public sidebarRoutes: any[] = [
-    { name: 'Overview', path: '/learn' },
+    { name: 'Overview', path: '/learn', children: [
+      { name: 'Test1', path: '/learn/test-1' },
+      { name: 'Test2', path: '/learn/test-2' }
+    ]},
```
Blackbaud-BrandonJones: Not sure we can just pass a children array to the current routes,  these may have to be checked and built?

Blackbaud-SteveBrush: I like this method. Need to make sure the `StacheNavLink` interface includes an option for `children?: StacheNavLink[]`.

blackbaud-brandonhare: YAS! SUCK IT JONES!

Blackbaud-BrandonJones: But... Can we count on the routes being passed in this way?   Aren't these routes coming from Angular2 / skyUX builder based on file structure?

Blackbaud-SteveBrush: That's the intention, eventually. Even still, that logic wouldn't exist on the component itself, but probably in the layout components. All nav components may now have "children", which will probably be more common than not...

src/modules/nav/nav.component.html
```diff
@@ -2,6 +2,11 @@
   <ol class="stache-nav-list">
     <li class="stache-nav-list-item" *ngFor="let route of routes">
       <a class="stache-nav-anchor" (click)="navigate(route)" [ngClass]="{'active': isActive(route)}">{{route.name}}</a>
+      <ol class="stache-nav-child-list" *ngIf="hasChildRoutes(route)">
```
Blackbaud-SteveBrush: I like the `hasChildRoutes` method. Possible to inject another `<stache-nav></stache-nav>` in here, instead of generating a new `ol`?

src/app/learn/learn.component.ts
blackbaud-brandonhare: That's what I was thinking. We can grab the raw routes object and return it in the form that the consuming components expect.

src/modules/nav/nav.component.html
blackbaud-brandonhare: You totally can! That's awesome!

src/modules/nav/nav.component.html
```diff
@@ -2,6 +2,9 @@
   <ol class="stache-nav-list">
     <li class="stache-nav-list-item" *ngFor="let route of routes">
       <a class="stache-nav-anchor" (click)="navigate(route)" [ngClass]="{'active': isActive(route)}">{{route.name}}</a>
+      <div class="stache-nav-child-list">
```
Blackbaud-SteveBrush: This DIV isn't really needed. We can target `.stache-nav .stache-nav` in the CSS.

src/modules/nav/nav.component.ts
```diff
@@ -24,6 +24,10 @@ export class StacheNavComponent implements OnInit, StacheNav {
     return (this.routes && this.routes.length > 0);
   }
 
+  public hasChildRoutes(route: any): boolean {
+    return (route.children && route.children.length > 0);
```
Blackbaud-SteveBrush: 👍

src/modules/nav/nav.component.scss
```diff
@@ -6,8 +6,12 @@
   flex-direction: column;
 }
 
+.stache-nav-list .stache-nav-list {
```
Blackbaud-SteveBrush: This should probably be placed in the sidebar SCSS file.

src/modules/nav/nav.component.html
```diff
@@ -2,6 +2,7 @@
   <ol class="stache-nav-list">
     <li class="stache-nav-list-item" *ngFor="let route of routes">
       <a class="stache-nav-anchor" (click)="navigate(route)" [ngClass]="{'active': isActive(route)}">{{route.name}}</a>
+      <stache-nav navType="sidebar" [routes]="route.children" *ngIf="hasChildRoutes(route)"></stache-nav>
```
Blackbaud-SteveBrush: 👍

src/modules/nav/nav.component.scss
blackbaud-brandonhare: But it targets the component itself?

Blackbaud-SteveBrush: Is this `padding-left` intended for sidebar navigations, though? For example, if we threw a child nav on the breadcrumbs, it would add padding to the left-hand side. `NavComponent` is designed to be bare-bones. If it makes more sense, we can use the `.stache-nav-sidebar` classes instead...

Blackbaud-SteveBrush:
@blackbaud-brandonhare @Blackbaud-BrandonJones This is great! We need to add a spec file to this PR, before we can merge.

src/modules/nav/nav.component.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+import { DebugElement } from '@angular/core';
+import { Router } from '@angular/router';
+import { expect } from '@blackbaud/skyux-builder/runtime/testing/browser';
+
+import { StacheNavComponent } from './nav.component';
+
+describe('StacheNavComponent', () => {
+  let component: StacheNavComponent;
+  let fixture: ComponentFixture<StacheNavComponent>;
+  let debugElement: DebugElement;
```
Blackbaud-SteveBrush: This isn't really getting used anywhere?

blackbaud-brandonhare: It was, but then I changed the test. I'll remove it.

PR #56, opened 2017-04-11
Style Adjustments

Blackbaud-SteveBrush:


blackbaud-brandonhare:
I still don't like the background color, but since we have the indentation, I'm cool with it.

PR #62, opened 2017-04-12
Migrated Action Buttons layout to a component

Blackbaud-SteveBrush:
- Component can now be used by itself, or within any layout.
- Action Buttons component extends Nav component


src/app/components/layout-sidebar/layout-sidebar-demo.component.html
```diff
@@ -1,3 +1,3 @@
-<stache-layout-sidebar>
+<stache layout="sidebar">
```
Blackbaud-SteveBrush: Changed this so that consumers can no longer use `<stache-layout-sidebar>`. They must use either `<stache layout="sidebar">` or `<stache-layout layoutType="sidebar">`. Eases confusion...

blackbaud-brandonhare: I think we should pick one way to do it and then document that. Even having those two options is confusing.

Blackbaud-SteveBrush: You're probably right. We'd essentially be requiring folks to use `<stache>` around everything, even for custom layouts.

PR #71, opened 2017-04-14
Feature sky window service

blackbaud-brandonhare:


src/app/app-extras.module.ts
```diff
@@ -15,6 +16,7 @@ require('style-loader!./app.scss');
     StacheModule
   ],
   providers: [
+    SkyAppConfig,
```
Blackbaud-SteveBrush: This is a bug, so not needed for this PR.

src/modules/back-to-top/back-to-top.component.spec.ts
```diff
@@ -1,10 +1,10 @@
 import { ComponentFixture, fakeAsync, inject, tick, TestBed } from '@angular/core/testing';
 import { DebugElement } from '@angular/core';
 import { expect } from '@blackbaud/skyux-builder/runtime/testing/browser';
+import { SkyAppWindowRef } from '@blackbaud/skyux-builder/runtime/window-ref';
```
Blackbaud-SteveBrush: You can import this directly from `@blackbaud/skyux-builder/runtime`.

PR #72, opened 2017-04-14
Added tests for page summary.

blackbaud-brandonhare:


src/app/app-extras.module.ts
```diff
@@ -15,6 +16,7 @@ require('style-loader!./app.scss');
     StacheModule
   ],
   providers: [
+    SkyAppConfig,
```
Blackbaud-SteveBrush: This is a bug, so not needed for this PR.

PR #91, opened 2017-04-18
Added tests for TestUtility. Moved to shared folder.

Blackbaud-SteveBrush:


src/modules/back-to-top/back-to-top.component.spec.ts
```diff
@@ -53,12 +53,12 @@ describe('StacheBackToTopComponent', () => {
 
   it('should show when the window y offset is greater than the specified offset', fakeAsync(() => {
     component.offset = 0;
-    TestUtility.fireDomEvent(windowRef, 'scroll');
+    TestUtility.triggerDomEvent(windowRef, 'scroll');
```
blackbaud-brandonhare: Did they rename that? Or do you like trigger better? Lol. I don't mind either way.

Blackbaud-SteveBrush: "Trigger" is more often used than "fire". [For example...](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events)

blackbaud-brandonhare: It makes sense. I literally copied that from SKYUX. I like triggered better. :)

PR #94, opened 2017-04-19
Updated tests for 100% branch coverage

Blackbaud-SteveBrush:


src/modules/code-block/code-block.component.ts
```diff
@@ -53,10 +53,8 @@ export class StacheCodeBlockComponent implements AfterViewInit {
     this.cdRef.detectChanges();
   }
 
-  public getClassNames(): string {
-    if (this.languageType) {
-      return `language-${this.languageType}`;
-    }
+  public getClassName(): string {
+    return `language-${this.languageType}`;
```
Blackbaud-BrandonJones: so if no languageType gets passed in,  it will return `language-undefined` ?   or is there a default?

blackbaud-brandonhare: There's a default. I think it's called 'default'

blackbaud-brandonhare: It returns `language-markup` which is the default.

PR #95, opened 2017-04-19
Test wrapper component

blackbaud-brandonhare:


src/modules/wrapper/wrapper.component.ts
```diff
@@ -43,14 +44,15 @@ export class StacheWrapperComponent implements OnInit, AfterContentInit {
...
     this.titleService.setTitle(this.windowTitle || this.pageTitle);
 
     this.route.fragment.subscribe(fragment => {
-      setImmediate(() => {
-        const element = document.getElementById(fragment);
+      this.windowRef.nativeWindow.setImmediate(() => {
+        const element = this.windowRef.nativeWindow.document.getElementById(fragment);
         if (element) {
           element.scrollIntoView();
         }
```
Blackbaud-SteveBrush: We need a newline at the end of the file.

src/modules/wrapper/wrapper.component.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+
+import { expect } from '@blackbaud/skyux-builder/runtime/testing/browser';
+import { SkyAppWindowRef } from '@blackbaud/skyux-builder/runtime';
+
+import { StacheWrapperComponent } from './wrapper.component';
+import { StacheLayoutComponent } from '../layout/layout.component';
+import { StacheWrapperTestComponent } from './fixtures/wrapper.component.fixture';
+import { StacheTitleService } from './title.service';
+
+describe('Stache Wrapper Component', () => {
```
Blackbaud-SteveBrush: Our other specs don't include spaces between the words `StacheWrapperComponent`.

src/modules/wrapper/wrapper.component.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    TestBed.configureTestingModule({
+      declarations: [
+        StacheLayoutComponent,
+        StacheWrapperComponent,
+        StacheWrapperTestComponent
+      ],
+      providers: [
+        { provide: ActivatedRoute, useValue: mockActivatedRoute },
+        { provide: Router, useValue: mockRouter },
```
Blackbaud-SteveBrush: Both `mockActivatedRoute` and `mockRouter` are not being used in the tests, so you can instead inject the class itself into the providers:

```
{ provide: ActivatedRoute, useClass: MockActivatedRoute }
```

karma.config.js
```diff
@@ -0,0 +1,67 @@
...
+      noInfo: true,
+      stats: 'minimal'
+    },
+    reporters: ['mocha', 'coverage'],
+    port: 9876,
+    colors: true,
+    logLevel: config.LOG_INFO,
+    autoWatch: false,
+    singleRun: true,
+    browserNoActivityTimeout: 100000
```
Blackbaud-SteveBrush: @blackbaud-brandonhare This line is what made it work. Apparently, the wrapper tests are taking quite a while and Chrome timed out. IMO we have two options: 1) Double-check the tests to make sure the code isn't caught in a time warp, or 2) submit a PR to Builder to add the `browserNoActivityTimeout` to the test config.

Blackbaud-SteveBrush: FYI, I added the karma.config.js file purely for debugging purposes.

Blackbaud-SteveBrush:
Also, code coverage is at 100%, but the branches are near 87.5%.

karma.config.js
blackbaud-brandonhare: If it's caught in a loop, why wouldn't it time out on our machines? That's strange. I believe it was working previously with `setImmediate`, but I don't want to go back to using a non standard window method if `setInterval` achieves the same result.

blackbaud-brandonhare: The other strange thing is that we are mocking `setTimeout`. It's not actually calling the native method. So I'm not sure why that would affect it.

package.json
```diff
@@ -10,7 +10,8 @@
     "node": ">=6.9.1"
   },
   "scripts": {
-    "skyux": "skyux"
+    "skyux": "skyux",
+    "test": "node --max-old-space-size=4096 ./node_modules/karma/bin/karma start ./karma.config.js --command test"
```
Blackbaud-SteveBrush: Sorry man, I didn't remove this from yesterday's "experiment". Mind?

src/modules/wrapper/wrapper.component.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+    component.pageTitle = 'Page Title';
+    fixture.detectChanges();
+    expect(mockTitleService.setTitle).toHaveBeenCalledWith('Page Title');
+  });
+
+  it('should grab the element from the fragment', async(() => {
+    component.ngOnInit();
+    fixture.detectChanges();
+    fixture.whenStable().then(() => {
+    expect(mockWindowService.nativeWindow.document.getElementById)
```
Blackbaud-SteveBrush: Needs a tab

PR #96, opened 2017-04-19
Auto-navigation for Breadcrumbs and Sidebar Components

Blackbaud-SteveBrush:
- Ignore styles for now
- Nav titles are derived from the directory name (i'll create an issue to derive the nav title from a `stache` tag attribute


src/modules/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -1,13 +1,60 @@
...
+    });
+
+    if (root.path !== '/') {
+      routes.push({
+        name: root.name,
+        path: root.path
+      });
+    }
+
+    const findActiveRoute = (r: any[]) => {
```
blackbaud-brandonhare: I know why you called this `r`, but can we name it something more descriptive? `interiorRoutes` or `routesArray` or something?

src/app/components/layout-sidebar/layout-sidebar-demo.component.ts
```diff
@@ -5,5 +5,19 @@ import { Component, OnInit } from '@angular/core';
   templateUrl: './layout-sidebar-demo.component.html'
 })
 export class StacheLayoutSidebarDemoComponent implements OnInit {
+  public routes: any[] = [
```
Blackbaud-BrandonJones: Does this have to be any?  Or do we have an interface of routes we can type this?

Blackbaud-SteveBrush: Ah, that's on the demo page... No we could pull in the `StacheNavLink` from the module, but I'd prefer not to require the consumer knowing which type to use for every attribute (different story when developing within the module).

src/modules/nav/nav.component.ts
```diff
@@ -59,6 +62,7 @@ export class StacheNavComponent implements OnInit, StacheNav {
     if (route.fragment) {
       extras.fragment = route.fragment;
       this.scrollToElement(route.fragment);
+      return;
```
blackbaud-brandonhare: Does this resolve the path issue for testing?

src/modules/nav/nav.service.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    let activeRoutes = navService.getActiveRoutes();
+    expect(activeRoutes[0].children.length).toBe(1);
+
+    skyAppConfig.runtime.routes = [];
+    activeRoutes = navService.getActiveRoutes();
+    expect(activeRoutes[0].children.length).toBe(1);
+  });
+
+  it('should only unset the active routes on NavigationStart', () => {
+    router.events = Observable.of(new NavigationEnd(0, '', ''));
```
blackbaud-brandonhare: wat

src/modules/nav/nav.component.ts
Blackbaud-SteveBrush: Not sure what you're referring to, sorry...

src/modules/nav/nav.service.spec.ts
Blackbaud-SteveBrush: Ha! Yeah, this was fun. It's the equivalent to ng1's router hooks. I wanted to clear the active routes when the user navigates. This test was the only way to reach a specific branch in the coverage.

src/modules/nav/nav.service.ts
```diff
@@ -0,0 +1,103 @@
...
+
+@Injectable()
+export class StacheNavService {
+  private activeRoutes: StacheNavLink[];
+
+  public constructor(
+    private appConfig: SkyAppConfig,
+    private router: Router
+  ) {
+    router.events.subscribe((val: any) => {
```
blackbaud-brandonhare: Cool!

src/modules/nav/nav.service.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  public getActiveRoutes(): StacheNavLink[] {
+    if (this.activeRoutes) {
+      return this.clone(this.activeRoutes) as StacheNavLink[];
+    }
+
+    let rootPath = this.getActiveUrl().replace(/^\//, '').split('/')[0];
+    let routes = this.clone(this.appConfig.runtime.routes);
+
+    routes = routes.filter((route: any) => route.routePath.indexOf(rootPath) === 0);
```
blackbaud-brandonhare: WHY YOU NOT CHAININ BRAH?

src/modules/nav/nav.service.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    routes = routes.map((route: any) => {
+      return {
+        segments: route.routePath.split('/'),
+        path: route.routePath
+      };
+    });
+
```
blackbaud-brandonhare: Really not a fan of the constant reassignment vs just chaining.

Blackbaud-SteveBrush: The methods (in this case) are doing very different things.

src/modules/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -1,13 +1,60 @@
...
+    });
+
+    if (root.path !== '/') {
+      breadcrumbRoutes.push({
+        name: root.name,
+        path: root.path
+      });
+    }
+
+    const findActiveRoute = (routes: any[]) => {
```
Blackbaud-BrandonJones: Is there a reason this method has to be defined as a const and live inside the filterRoutes method?  Would it not be better to have it live separate?

src/modules/nav/nav.component.ts
```diff
@@ -22,7 +22,7 @@ export class StacheNavComponent implements OnInit, StacheNav {
 
   public constructor(
     private router: Router,
-    private windowRef: SkyAppWindowRef) {}
+    private windowRef: SkyAppWindowRef) { }
```
Blackbaud-BrandonJones: is the space between empty curly brackets going to be our syntax standard?

src/modules/nav/nav.service.spec.ts
blackbaud-brandonhare: Those empty string parameters. What a weird API.

src/modules/nav/nav.service.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  private toTitleCase(phrase: string): string {
+    return phrase
+      .split(' ')
+      .map(word => word.charAt(0).toUpperCase() + word.slice(1))
+      .join(' ');
+  }
+
+  private clone(thing: any): any {
```
blackbaud-brandonhare: Thing? :P Are you cloning anything other than a route object? Can we be more prescriptive about the name?

src/modules/nav/nav.service.ts
blackbaud-brandonhare: That's the case with every method! They always do different things. We could create named functions for those in the chain for clarity. As is, this is still not super clear and readable IMO.

blackbaud-brandonhare:
```
let childRoutes = this.clone(this.appConfig.runtime.routes)
    .filter((route: any) => route.routePath.indexOf(rootPath) === 0)
    .map((route: any) => {
        return {
            segments: route.routePath.split('/'),
            path: route.routePath
        };
    });

let routes = [{
    path: rootPath,
    segements: [rootPath],
    children: this.orderRoutes(childRoutes, rootPath);
}];
```

Blackbaud-BrandonJones: This is a lot cleaner, though I would prefer not to chain off a method on our class.   

```
let clonedRoutes = this.clone(this.appConfig...)
let childRoutes = clonedRoutes.filter(....
```
Chain array methods off an array, not off the data returned from our method.  Especially since `this.clone` can return an object, not just an array.

blackbaud-brandonhare: I think this is semantics, but this could be an example of clearer naming conventions. The clone method should be `cloneRoutes(routes: [])` so we would know what we were getting off that. This just adds an extra step of assignment and doesn't add to the clarity IMO.

blackbaud-brandonhare: In any case, the status quo of reassigning that variable over and over is obtuse.

src/modules/nav/nav.component.ts
Blackbaud-SteveBrush: Some linters require spaces between curly brackets. This is a convention I've seen on many Angular 2 docs.

src/modules/nav/nav.service.ts
Blackbaud-SteveBrush: In this case, the method is simply cloning a thing (could be a string, an array, or whatever).

src/modules/nav/nav.service.ts
Blackbaud-BrandonJones: but the versatility of the clone method is interesting,  it could be used to clone objects or arrays in the future.  DRY, though I suppose non-typescripty.      can we type variables?   `let clonedRoutes: routes[] = ` ?

src/modules/breadcrumbs/breadcrumbs.component.ts
Blackbaud-SteveBrush: Yep; it's directly modifying an array in its parent context.

src/modules/nav/nav.service.ts
blackbaud-brandonhare: But this is a small scoped service. Why do we need to design with a bunch of not-needed hypotheticals? Make it simple, make it prescriptive. If we need to be able to clone a bunch of different stuff, then we could revisit.

src/modules/nav/nav.component.ts
Blackbaud-BrandonJones: I like it

src/modules/nav/nav.service.ts
blackbaud-brandonhare: But I would also be ok with being more descriptive with the variable names with typing too. That would still be more readable.

Blackbaud-BrandonJones: I guess if it were `cloneRoutes(routes: []): route[]`  chaining would be fine off of it.  But we would lose the versatility of that method.    I am good with either way here,  being more specific with that method and chaining, or calling the method separately getting the data and then calling the array methods on it.

src/modules/sidebar/sidebar.component.scss
```diff
@@ -17,24 +26,33 @@
...
-  .stache-nav-sidebar .stache-nav-sidebar {
-    background-color: $stache-color-gray-light;
-    padding: 0; 
-    margin-bottom: 8px;
+  .stache-nav-sidebar {
+    .stache-nav-sub {
+      background-color: $stache-color-gray-light;
+      padding: 0; 
+      margin-bottom: 8px;
+      display: none;
```
blackbaud-brandonhare: I'm assuming you're showing this programmatically somewhere?

src/modules/sidebar/sidebar.component.ts
```diff
@@ -1,13 +1,34 @@
...
+    if (!this.routes) {
+      const activeRoutes = this.navService.getActiveRoutes();
+      this.routes = this.filterRoutes(activeRoutes);
+    }
+  }
+
+  private filterRoutes(activeRoutes: StacheNavLink[]): StacheNavLink[] {
+    const root = activeRoutes[0];
+
+    root.children.unshift({
```
blackbaud-brandonhare: So we are creating an overview route by default for every sidebar link? What if we don't want this or we want to change the name? I suppose this works for now, but in the future we might want to allow for setting a 'default' starting route perhaps?

src/modules/nav/nav.component.ts
```diff
@@ -33,17 +33,20 @@ export class StacheNavComponent implements OnInit, StacheNav {
...
-    }
-
-    if (this.router.url.includes('#')) {
-      return (this.router.url.split('#')[0] === path);
     } else {
-      return (this.router.url === path);
+      navDepth = path.split('/').length;
     }
+
+    const isActiveParent = (navDepth > 2  && `${activeUrl}/`.indexOf(`${path}/`) === 0);
```
blackbaud-brandonhare: Yeah, I would recommend making that active check a method on the service. Would look a lot cleaner. :)

src/modules/sidebar/sidebar.component.scss
Blackbaud-SteveBrush: Well, the css is handling the display of the submenus. As it stands, all submenus are available, but only the active submenus are visible to the end user. This should give folks flexibility to show those other navs if needed.

src/modules/sidebar/sidebar.component.ts
Blackbaud-SteveBrush: Yeah, I wrestled with this. We can see how consumers react to it...

src/modules/nav/nav.component.ts
Blackbaud-SteveBrush: On second look, the context for checking if a link is active is different, so a service wouldn't benefit us much just yet.

src/modules/nav/nav.service.ts
```diff
@@ -0,0 +1,105 @@
...
+          path: route.routePath
+        };
+      });
+
+    let activeRoutes = [{
+      path: rootPath,
+      segments: [rootPath],
+      children: this.orderRoutes(activeChildRoutes, rootPath)
+    }];
+
```
Blackbaud-BrandonJones: 👍

src/modules/nav/nav.component.ts
blackbaud-brandonhare: 👍

src/modules/nav/nav.service.ts
blackbaud-brandonhare: MUCH better. Thanks Steve!

src/modules/sidebar/sidebar.component.scss
blackbaud-brandonhare: Cool. I figured, I just didn't see where that CSS was manipulated in the sidebar nav component.

src/modules/sidebar/sidebar.component.ts
blackbaud-brandonhare: Hmm. Yeah, I still don't love it, but we can see how it goes. I wish I could test it locally and see how it actually renders...

PR #99, opened 2017-04-20
Unit Tests > InputConverter

Blackbaud-SteveBrush:


src/modules/shared/input-converter.ts
```diff
@@ -1,31 +1,41 @@
-const stringConverter = (value: any) => {
+import {} from 'reflect-metadata';
```
blackbaud-brandonhare: This already just looks cleaner from a glance.

src/modules/shared/input-converter.ts
```diff
@@ -1,31 +1,41 @@
...
   }
 
   return parseFloat(value.toString());
 };
 
+const getMetaData = (target: Object, key: string) => {
+  return Reflect.getMetadata('design:type', target, key);
+};
+
 export function InputConverter(converter?: (value: any) => any) {
```
blackbaud-brandonhare: Yeah, exporting functions makes them really difficult to spy on. We need to remember this for the future. We should probably make it a design pattern.

PR #102, opened 2017-04-21
Code coverage for layouts.

blackbaud-brandonhare:


src/modules/layout/layout-blank.component.spec.ts
```diff
@@ -0,0 +1,54 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { NO_ERRORS_SCHEMA } from '@angular/core';
```
Blackbaud-SteveBrush: You throw this in here because the tests were slow, or just because?

src/modules/layout/layout.component.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    fixture.detectChanges();
+    let layout = component['blankTemplateRef'];
+    expect(component.templateRef === layout).toBe(true);
+  });
+
+  it('should set the template ref to default given the layoutType', () => {
+    component.layoutType = 'default';
+    fixture.detectChanges();
+    let layout = component['defaultTemplateRef'];
+    expect(component.templateRef === layout).toBe(true);
```
Blackbaud-SteveBrush: why not just `expect(component.templateRef).toBe(layout);`?

src/modules/layout/layout-blank.component.spec.ts
blackbaud-brandonhare: I thought it was for templates at first, but then realized the error was something else. I'll remove it.

src/modules/layout/layout.component.spec.ts
blackbaud-brandonhare: That works too.  6 of one, 1/2 dozen of the other.

src/modules/layout/layout-blank.component.spec.ts
blackbaud-brandonhare: Nope, you absolutely need it for the nested components.

PR #108, opened 2017-04-25
Styles

Blackbaud-SteveBrush:


src/modules/layout/container.component.scss
```diff
@@ -1,3 +1,22 @@
...
+
 .stache-container {
   max-width: 100%;
+}
+
+/deep/
+.stache-container {
+  p,
+  ul,
+  ol,
```
blackbaud-brandonhare: It...just...keeps...going! :P

src/modules/page-anchor/page-anchor.component.spec.ts
```diff
@@ -53,10 +53,10 @@ describe('StachePageAnchorComponent', () => {
     expect(id).toBe('test-content');
   });
 
-  it('should call the addHashToUrl method when the heading is clicked', () => {
+  it('should call the addHashToUrl method when the icon is clicked', () => {
```
blackbaud-brandonhare: No more clicking on the header itself to scroll to it? I suppose that works too.

PR #124, opened 2017-05-01
Removed app files except shared

blackbaud-brandonhare:


Blackbaud-BrandonJones:
If this is the direction we are 100% sure we want to go... then I say sweet.

Blackbaud-BrandonJones:
Actually, maybe we should just remove the components folder?  Keeping the home page/about to use while creating components?    As we are only going to be bundling and exporting the stache file, I don't think it would hurt to keep some front-end matter in this project for dev purposes.

blackbaud-brandonhare:
Yeah @Blackbaud-BrandonJones that's why I put that `index.html` file back in there. That's really all you need for testing purposes.

Blackbaud-BrandonJones:
Unless you're testing routes and layouts :)

blackbaud-brandonhare:
Who needs to do that?

PR #126, opened 2017-05-02
Video Component

blackbaud-brandonhare:
Responsive video embed component.


Blackbaud-BrandonJones:
From what little I do know about video embedding, this does look good to me.  @Blackbaud-SteveBrush ?

src/app/shared/stache/modules/video/video.component.html
```diff
@@ -0,0 +1,3 @@
+<div class="stache-video">
+  <iframe class="stache-video-embed" [src]="src" frameborder="0" allowfullscreen></iframe>
```
Blackbaud-SteveBrush: Small change, but can we use `src="{{videoSource}}"` so that the consumer is not confused about the input? `src` is owned by `iframe`, and isn't an angular input property.

src/app/shared/stache/modules/video/video.component.scss
```diff
@@ -0,0 +1,23 @@
+@import '../../styles/_variables';
+
+/deep/
```
Blackbaud-SteveBrush: `/deep/` not needed, here.

src/app/shared/stache/modules/video/video.component.ts
```diff
@@ -0,0 +1,22 @@
...
+  templateUrl: './video.component.html',
+  styleUrls: ['./video.component.scss']
+})
+export class StacheVideoComponent implements OnInit {
+  @Input()
+  public videoSource: string;
+
+  public src: SafeResourceUrl;
+
+  constructor(
```
Blackbaud-SteveBrush: Just to keep things consistent with other components...

```
constructor(
  private sanitizer: DomSanitizer) { }
```

src/app/shared/stache/modules/video/video.component.html
Blackbaud-SteveBrush: Just thinking out loud: this component is literally just an iframe; there's nothing "video" about it. Couldn't we just call this a `<stache-iframe>`?

Blackbaud-SteveBrush: I _do_ like having a `<stache-video>` component; just thinking about future implementations. If someone wanted an `iframe`, we'd have to create another component that looks very similar to this one. We can come back to this discussion in the future; just some thoughts.

blackbaud-brandonhare: The consumer wouldn't ever see this though? But I can do that too.

blackbaud-brandonhare: Hmm. Yeah I suppose we could make it more multi-purpose. But the styles are specifically tailored to video embeds. What if I created a `stache-iframe` component and used it here instead? Then we could have both?

Blackbaud-SteveBrush: We can come back to this idea later, just wanted to get our minds thinking about it. I believe SKY UX is planning on building an iframe component.

blackbaud-brandonhare: Security doesn't work the other way. It needs to be bound like that.

blackbaud-brandonhare: Sounds good!

Blackbaud-SteveBrush: Last comment, then this can be merged: 
- Move the closing iframe tag to a new line
- Rename class to `stache-video-object` to be consistent with the image component

This looks good!

PR #133, opened 2017-05-03
Includes Component

blackbaud-brandonhare:
https://github.com/blackbaud/skyux-builder-plugin-stache-include


Blackbaud-SteveBrush:
Thoughts on a singular "include"? `<stache-include source=""></stache>` Since we're only including the one file...

Blackbaud-BrandonJones:
I like that better personally.   Not a fan of `fileName=""` for some reason,  though I also think that `source` might be a bit confusing and users may instinctually feel they need to put the relative path to the file instead of just the name of it. 
Maybe something like `<stache-include includeTemplate="foo.html"></stache-include>

Blackbaud-SteveBrush:
We can use this PR to also talk about the plugin source. 
- Do you mind writing tests (and implementation) for when the file requested does NOT exist?
- My preference would be to include an error message in the terminal during the build process.

blackbaud-brandonhare:
Updated plugin source @Blackbaud-SteveBrush

Blackbaud-SteveBrush:
Go ahead and add an entry for the plugin in **skyuxconfig.json**. Got a PR in place to do the same for the `skyux-template-stache`?

blackbaud-brandonhare:
@Blackbaud-SteveBrush 
1. Done
2. Yes https://github.com/blackbaud/skyux-template-stache/pull/11

PR #147, opened 2017-05-11
Making some suggestions

Blackbaud-SteveBrush:


config/webpack/webpack.bundle.config.js
```diff
@@ -7,9 +7,9 @@ module.exports = {
   entry: helpers.root('.srctmp/index.ts'),
   output: {
     path: path.resolve(__dirname, '..', '..', 'dist', 'bundles'),
-    filename: 'core.umd.js',
+    filename: 'stache.umd.js',
```
Blackbaud-SteveBrush: I used "core" so that any library can use it. Are you assuming some level of configuration, here, for builder version?

Blackbaud-SteveBrush: ie, when typing `skyux release` (or something), builder would take the app's name from skyuxconfig.json?

Blackbaud-BobbyEarl: Yeah I think that's something to keep in mind as configurable when we migrate to builder.

PR #148, opened 2017-05-11
Video Component suggestions...

Blackbaud-SteveBrush:


src/app/index.html
```diff
@@ -44,7 +44,15 @@
   </stache-page-anchor>
 
   <stache-image
-    imageSource="https://placecage.com/200/200">
+    imageSource="https://unsplash.it/200/300">
```
Blackbaud-SteveBrush: Need SSL support, plus this is boring and plain :)

src/app/public/src/modules/video/video.component.scss
```diff
@@ -3,19 +3,30 @@
 .stache-video {
   position: relative;
   display: block;
-  height: 0;
-  padding: 0;
-  padding-bottom: 56.25%;
+  width: 100%;
   overflow: hidden;
-  margin: 2em 0;
+  margin-top: 1em;
```
Blackbaud-SteveBrush: `1em` is what's generally used for block elements throughout the app.

src/app/index.html
blackbaud-brandonhare: LAAAAAME.

src/app/public/src/modules/video/video.component.scss
blackbaud-brandonhare: Yeah, I was gonna change that after the Image PR. Makes sense to be consistent.

PR #160, opened 2017-05-16
Feature global variables

blackbaud-brandonhare:


src/app/index.html
```diff
@@ -15,7 +15,8 @@
   pageTitle="Welcome"
   layout="container"
   showBreadcrumbs="true"
-  showTableOfContents="true">
+  showTableOfContents="true"
+  #stache>
```
Blackbaud-SteveBrush: Can you explain what adding `#stache` does?

Blackbaud-SteveBrush: The reason I ask: I'm not sure what we gain, from just simply documenting `{{ config.skyux.appSettings.stache.example }}` (or even `{{ config.skyux.appSettings.example }}`).

The confusion of adding `#stache` to the `<stache>` tag isn't worth the small benefit of a simpler object chain, in my opinion.

blackbaud-brandonhare: The `#stache` gives the template access to the component's variables. The benefit we get is reinforcing a common pattern in stache and keeping it simple. Also, Bobby recommended we use a stache namespace in the config to prevent potential collision with other things.

blackbaud-brandonhare: Also, we could add a plugin to add the `#stache` attribute to the `<stache>` component tags if that's a concern.

Blackbaud-SteveBrush:
Closing because we'll be putting this in a service.

Issue #158, opened 2017-05-15
Increase line-heights of Table Of Contents Component


Blackbaud-SteveBrush:
![pasted_image_at_2017_05_15_03_48_pm_360](https://cloud.githubusercontent.com/assets/12497062/26076316/5a12e438-3986-11e7-8262-a6167cc554a8.png)

PR #164, opened 2017-05-17
Adjusted nav line-height

Blackbaud-SteveBrush:
Addresses #158 and #176.


src/app/public/src/modules/layout/container.component.scss
```diff
@@ -19,4 +19,18 @@
     line-height: $stache-line-height;
     color: $stache-color-text;
   }
+
+  ul:not([class]),
```
blackbaud-brandonhare: How does this work? I'm curious.

PR #166, opened 2017-05-18
Ability to change route name using `pageTitle`, `navTitle`

Blackbaud-SteveBrush:
- Requires [skyux-builder-plugin-stache-route-metadata](https://github.com/blackbaud/skyux-builder-plugin-stache-route-metadata).
- Need to think about how we can also manage route order using this method (in a different PR)


package.json
```diff
@@ -33,11 +33,12 @@
     "@blackbaud/skyux": "2.0.0-beta.27",
     "@blackbaud/skyux-builder-plugin-stache-code-block": "1.0.0",
     "@blackbaud/skyux-builder-plugin-stache-include": "1.0.0",
+    "@blackbaud/skyux-builder-plugin-stache-route-metadata": "blackbaud/skyux-builder-plugin-stache-route-metadata#unit-tests",
```
Blackbaud-BrandonJones: are we pointing to #unit-tests on purpose?

package.json
```diff
@@ -33,11 +33,12 @@
...
     "prismjs": "1.6.0",
     "smoothscroll-polyfill": "0.3.5"
   },
   "devDependencies": {
-    "@blackbaud/skyux-builder": "1.0.0-beta.25",
+    "@blackbaud/skyux-builder": "blackbaud/skyux-builder#plugin-config",
```
Blackbaud-BrandonJones: also plugin-config?

package.json
Blackbaud-SteveBrush: Yeah, until y'all approve that PR, too :)

package.json
Blackbaud-SteveBrush: This can be updated. One minute...

src/app/public/src/modules/shared/route-metadata.service.ts
```diff
@@ -0,0 +1,5 @@
+// The contents of this file are automatically generated by
```
blackbaud-brandonhare: Good use of comments!

Blackbaud-SteveBrush:
https://github.com/blackbaud/stache2/issues/168

PR #175, opened 2017-05-24
Hero > Moved image to below content

Blackbaud-SteveBrush:


src/app/public/src/modules/hero/hero.component.html
```diff
@@ -1,9 +1,11 @@
...
   </div>
   <div class="stache-hero-overlay"
     [ngStyle]="{ 'opacity': overlayOpacity }">
   </div>
-</div>
\ No newline at end of file
+  <div class="stache-hero-image"
+    [ngStyle]="{ 'background-image': 'url(' + backgroundImageUrl + ')' }">
+  </div>
+</div>
```
Blackbaud-BrandonJones: Why was this necessary?

PR #180, opened 2017-05-30
Refactored Affix Directive

Blackbaud-SteveBrush:
This is my fault: after running the code through some tests, I found a few issues that were a bit too difficult to describe.

I wanted the ability to add the directive to both native elements (`div`, `p`, etc.) and other components. Turns out that there is a slight difference in the two and how Angular handles directives.

Also, I streamlined the number of calculations being made every time the window is scrolled. 

Curious what your thoughts are, @Blackbaud-BrandonJones.


Blackbaud-BrandonJones:
I like it.   I don't know what some of those things are doing, but sure, if it works let's just move along.

Blackbaud-BrandonJones:
Some weird issues when re-sizing the browser and with "popping" when pulling an element out of the page context.   You can see with the Alert,  slowly scroll the the top, and it "pops" to the top, and the page around it jumps to fill in the gap.    Also when resizing the browser, and scrolling, it no longer respects the top of the browser, instead it keeps the original Y offset value.    If this isn't a concern then it's fine by me.

src/app/public/src/modules/shared/affix-top.directive.ts
```diff
@@ -1,39 +1,100 @@
...
   }
 
-  // private setElementStyle(property: string, value: string) {
-  //   // TODO: Will have to be updated to the renderer2 when we upgrade to Angular 4.x.
-  //   this.renderer.setElementStyle(this.elementRef.nativeElement, property, value);
-  // }
+  private isComponent(element: any): boolean {
+    let isComponent = false;
+
+    Array.prototype.slice.call(element.attributes)
```
blackbaud-brandonhare: es 2015 has a nice method so you don't have to do this anymore. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/of

Blackbaud-SteveBrush:
@blackbaud-brandonhare @Blackbaud-BrandonJones This needs another look. I added a component (that uses the directive) that replaces the affixed item with an invisible, matching-height `div`. This will allow the page to not "jump" when the item is affixed.

PR #185, opened 2017-05-31
Stache Analytics

Blackbaud-BrandonJones:
on hold to test and make sure it works.


Blackbaud-SteveBrush:
Tests are failing, but the Travis build passes. We need to look into why that is! #193

Blackbaud-SteveBrush:
I know you said you're working on tests. This is an FYI:

screen shot 2017-06-01 at 9 43 34 pm

Blackbaud-SteveBrush:
So far this is looking really good :) I'd rename the directory and files to `analytics`, instead of `stache-analytics`, just to keep it consistent with the other modules.

skyuxconfig.json
```diff
@@ -9,5 +9,10 @@
   ],
   "app": {
     "title": "Stache 2"
+  },
+  "appSettings": {
+    "stache": {
+      "googleAnalyticsClientId": "UA-2418840-1"
```
blackbaud-brandonhare: This is a good use case for `appSettings`

src/app/public/src/modules/analytics/google-analytics.component.ts
```diff
@@ -0,0 +1,42 @@
...
+    private windowRef: StacheWindowRef,
+    private configService: StacheConfigService) {
+      if (this.configService.runtime.command === 'build') {
+        this.init();
+      }
+    }
+
+  public init(): void {
+    const clientId = this.configService.skyux.appSettings.stache.googleAnalyticsClientId;
+    this.windowRef.nativeWindow.eval(`
```
blackbaud-brandonhare: @Blackbaud-SteveBrush isn't `eval` a HUGE no no?

src/app/public/src/modules/analytics/google-analytics.component.ts
```diff
@@ -0,0 +1,42 @@
...
+          (i[r].q=i[r].q||[]).push(arguments)},
+          i[r].l=1*new Date();
+      })(window,'ga');
+    `);
+
+    this.windowRef.nativeWindow.ga('create', clientId, 'auto');
+
+    this.router.events.subscribe(event => {
+      if (event instanceof NavigationEnd) {
+        this.windowRef.nativeWindow.ga('set', 'page',
```
blackbaud-brandonhare: This handles the page views? This is cool!

src/app/public/src/modules/analytics/google-analytics.component.ts
Blackbaud-SteveBrush: Yep, if you're not careful :) Angular uses it everywhere. Basically, `eval` is appropriate when the content you are evaluating is either static, or "known". https://stackoverflow.com/questions/197769/when-is-javascripts-eval-not-evil

src/app/public/src/modules/analytics/analytics.component.html
```diff
@@ -0,0 +1,4 @@
+<!-- Google Tag Manager -->
+<noscript><iframe src="//www.googletagmanager.com/ns.html?id=GTM-W56QP9"
```
Blackbaud-SteveBrush: This will never run. To test: disable JavaScript in your browser and type `skyux serve`. The `<noscript>` tag won't be added to the markup.

We can probably kill this template altogether and make this into a directive `stacheGoogleAnalytics`.

Blackbaud-BrandonJones: It still adds the iframe to the tag doesn't it?

Blackbaud-SteveBrush: This template doesn't even see the light of day if JS is disabled.

skyuxconfig.json
```diff
@@ -11,5 +11,10 @@
   ],
   "app": {
     "title": "Stache 2"
+  },
+  "appSettings": {
+    "stache": {
```
Blackbaud-SteveBrush: It would probably be better to isolate the configuration under `"googleAnalytics"`. Also, the user should be able to enable/disable analytics manually (defaults to `true`). The use case would be that they wish to disable analytics for a short period of time, but maintain their credentials in the config.

```
{
  "stache": {
    "googleAnalytics": {
      "enabled": "true",
      "analyticsClientId": "",
      "tagManagerContainerId": ""
    }
  }
}
```

src/app/public/src/modules/analytics/analytics.component.ts
```diff
@@ -0,0 +1,68 @@
...
+})
+export class StacheAnalyticsComponent implements OnInit {
+   public constructor(
+    private windowRef: StacheWindowRef,
+    private configService: StacheConfigService,
+    private router: Router) { }
+
+  public ngOnInit(): void {
+    let gtmScript = this.windowRef.nativeWindow.document.getElementById('stache-gtm-script');
+    if (this.configService.runtime.command === 'build' && !gtmScript) {
```
Blackbaud-SteveBrush: We could simply check for `ga` on the `window`, yes?
```
const isLoaded = this.windowRef.nativeWindow.ga;
const isProduction = this.configService.runtime.command === 'build';
if (isProduction && !isLoaded) { }
```

src/app/public/src/modules/analytics/analytics.component.ts
```diff
@@ -0,0 +1,68 @@
...
+          (i[r].q=i[r].q||[]).push(arguments)},
+          i[r].l=1*new Date();
+      })(window,'ga');
+    `);
+    this.windowRef.nativeWindow.ga('create', clientId, 'auto');
+  }
+
+  public bindPageViewsToRouter(): void {
+    this.router.events.subscribe(event => {
+      /* istanbul ignore else*/
```
Blackbaud-SteveBrush: Any specific reason why you're ignoring this test? I did a similar test, here: https://github.com/blackbaud/stache2/blob/master/src/app/public/src/modules/nav/nav.service.spec.ts#L81

src/app/public/src/modules/analytics/analytics.module.ts
```diff
@@ -0,0 +1,19 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { RouterModule } from '@angular/router';
+
+import { StacheAnalyticsComponent } from './analytics.component';
+
+@NgModule({
+  imports: [
+    CommonModule,
```
Blackbaud-SteveBrush: `CommonModule` isn't needed.

src/app/public/src/modules/analytics/analytics.component.ts
Blackbaud-BrandonJones: because there is no reason to test the else statement here.   We aren't writing the router event subscribe method.  or passing in a parameter for the if statement, so theres no reason to test the else/if statement here.  just that this method gets run.

PR #194, opened 2017-06-02
Updated stache to use new data plugin.

blackbaud-brandonhare:


Blackbaud-SteveBrush:
**Plugin should:**
- Handle and parse various file names
- Run unit tests

Blackbaud-BrandonJones:
Update the page-anchor component to use AfterContentChecked instead of ngOnInit, so if we add a variable to a page anchor it will still render in the table of contents.

src/app/public/src/modules/page-anchor/page-anchor.component.ts
```diff
@@ -19,7 +19,7 @@ export class StachePageAnchorComponent implements OnInit, StacheNavLink {
     private elementRef: ElementRef,
     private windowRef: StacheWindowRef) { }
 
-  public ngOnInit(): void {
+  public ngAfterContentChecked(): void {
```
Blackbaud-SteveBrush: Curious about why this was changed?

blackbaud-brandonhare: This needed to be changed in order to be able to use variables in the page anchor component. With the way it was before using `ngOnInit`, the content would render, but the TOC would be blank. It's a bug that Lindsey filed a while back.

src/app/public/src/modules/shared/json-data.service.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+    expect(data.global.productNameLong).toBe('Stache 2');
+  });
+
+  it('should return data from a specific key', () => {
+    const dataService = new StacheJsonDataService();
+    let data = dataService.getByKey('global');
+    expect(data.productNameLong).toBe('Stache 2');
+  });
+
+  it('should return undefined if the key does not exist', () => {
```
blackbaud-brandonhare: Ah, good catch!

src/app/public/src/modules/page-anchor/page-anchor.component.ts
Blackbaud-SteveBrush: Boyscoutin' it. Awesome!

Blackbaud-BrandonJones: 👍

PR #242, opened 2017-06-23
Feature route ordering

Blackbaud-BrandonJones:
dependent on https://github.com/blackbaud/skyux-builder-plugin-stache/pull/10


src/app/public/src/modules/nav/nav-link-metadata.ts
```diff
@@ -0,0 +1,4 @@
+export interface StacheNavLinkMetadata {
```
blackbaud-brandonhare: I really like interfaces. Probably my favorite part of TS.

src/app/public/src/modules/nav/nav.service.ts
```diff
@@ -45,19 +42,27 @@ export class StacheNavService {
...
     this.activeRoutes = undefined;
   }
 
-  private orderRoutes(routes: any[], parentPath: string): any[] {
+  private clone(thing: any): any {
+    return JSON.parse(JSON.stringify(thing));
+  }
+
+  private assignChildren(routes: any[], parentPath: string): any[] {
     const ordered: any[] = [];
```
Blackbaud-SteveBrush: Can you think of a better name than `ordered`, since the name of the function has changed? Maybe `branch`?

src/app/public/src/modules/nav/nav.service.ts
```diff
@@ -120,7 +129,37 @@ export class StacheNavService {
...
   }
 
-  private clone(thing: any): any {
-    return JSON.parse(JSON.stringify(thing));
+  private sortRoutes(routes: StacheNavLink[]): StacheNavLink[] {
+
+    const sortByName = (a: any, b: any): number  => {
+      if (a.name.toLowerCase() < b.name.toLowerCase()) {
+        return -1;
+      } else if (a.name.toLowerCase() > b.name.toLowerCase()) {
```
Blackbaud-SteveBrush: `else if` is redundant because you're using `return`.
```
if (a.name.toLowerCase() < b.name.toLowerCase()) {
  return -1;
}

if (a.name.toLowerCase() > b.name.toLowerCase()) {
  return 1;
}
```

Blackbaud-BrandonJones: That's correct, however to get tests/lint to go over all the "branch" options it gets frustrated with if statements without else/else ifs.    Also, the `sort` method whenever you look them up, all seem to be written in this format.   #reasonsTestsRuinCode

Blackbaud-SteveBrush: Ah I see. Thanks for clarifying!

src/app/public/src/modules/nav/nav.service.ts
```diff
@@ -76,36 +81,40 @@ export class StacheNavService {
...
+      let foundRoute = allMetadata.filter((metaRoute: any) => {
+        return metaRoute.path === route.path;
       })[0];
 
       if (foundRoute) {
-        return foundRoute.name;
+        if (foundRoute.order !== undefined) {
+          foundRoute.order = parseInt(foundRoute.order, 10);
+        }
+        pathMetadata = Object.assign({}, pathMetadata, foundRoute);
```
Blackbaud-SteveBrush: I might be missing something; is this `Object.assign` unnecessary? `pathMetadata` will always be `{}` at this moment.

Blackbaud-BrandonJones: Ah no you are correct.   I forgot to remove this with my last rework, good catch.

PR #245, opened 2017-06-29
Added more language support.

blackbaud-brandonhare:


Blackbaud-SteveBrush:
Do you mind adding a simple code block example for each language type to the Home page? This way we can verify that every type is covered.

src/app/public/src/modules/code-block/code-block.component.ts
```diff
@@ -3,8 +3,29 @@ import { DomSanitizer, SafeHtml } from '@angular/platform-browser';
...
+import 'prismjs/components/prism-powershell';
+import 'prismjs/components/prism-python';
+import 'prismjs/components/prism-rest';
+import 'prismjs/components/prism-ruby';
+import 'prismjs/components/prism-sass';
+import 'prismjs/components/prism-scss';
+import 'prismjs/components/prism-sql';
+import 'prismjs/components/prism-swift';
+import 'prismjs/components/prism-typescript';
+import 'prismjs/components/prism-yaml';
```
Blackbaud-BrandonJones: They don't expose a module or some other method of bundling all these at once we could use instead of each one individually?

Blackbaud-BrandonJones: http://prismjs.com/plugins/autoloader/ can't that load in the component grammars from the node_modules file?

blackbaud-brandonhare: we still have to import them anyway. Not sure what that buys us.

Blackbaud-BrandonJones:
```
Prism.plugins.autoloader.languages_path = 'path/to/grammars/in/node_modules';
```
They are already there, why can't the plugin handle loading them all up?

blackbaud-brandonhare: That might work if we had control over the bundling and build. But the import is what pulls all of that stuff in. Unless we want to load all of those in via the externals in `skyuxconfig`. In which case, we might as well just leave them here.

src/app/home.component.html
```diff
@@ -63,6 +63,236 @@
...
+  <stache-code-block
+    languageType="yaml">
+    ---
+    layout: layout-base
+    order: 0
+    name: Name of Page
+    description: This is a description
+    ---
+  </stache-code-block>
+
```
Blackbaud-BrandonJones: All these code block samples seems a bit... excessive and cluttering of the home page.   Wouldn't a list of supported language types under the code block component suffice?

PR #262, opened 2017-07-12
Added external links to action buttons.

blackbaud-brandonhare:


src/app/public/src/modules/nav/nav.component.ts
```diff
@@ -41,6 +41,11 @@ export class StacheNavComponent implements OnInit, StacheNav {
   public navigate(route: any): void {
     let extras: any = {};
 
+    if (route.isExternalLink) {
```
Blackbaud-SteveBrush: Would something like this work?

```
const isExternalUrl = (typeof route.path === 'string' && route.path.match(/^http/));
if (isExternalUrl) {
  // redirect...
}
```

Blackbaud-BrandonJones: ^---

blackbaud-brandonhare: Yeah I was on the fence about this anyway. I can just do the regex. I think that should cover us in most cases. I originally went with the boolean to account for any weirdness in URLs or different types of URLS. That could potentially be a mailto link or something, for example.

blackbaud-brandonhare: I suppose we could just update the regex to account for that if need be.

Blackbaud-SteveBrush: That's a good point. I'd go ahead and throw in the `/^mailto:/` regex as well!

blackbaud-brandonhare: #PairProgramming

blackbaud-brandonhare: Here's another thought. What if one of the routes contains http for some reason? Like an http topic?

Blackbaud-BrandonJones: you can regex to make it only account for the start of the string....

blackbaud-brandonhare: UGH FINE. :P

src/app/public/src/modules/nav/nav.component.spec.ts
```diff
@@ -101,6 +106,10 @@ describe('StacheNavComponent', () => {
     expect(route.isCurrent).toBe(true);
   });
 
+  it('should return true if a given route is external', () => {
+
+  });
```
Blackbaud-BrandonJones: ? empty test case?

src/app/public/src/modules/nav/nav.component.ts
```diff
@@ -106,4 +111,13 @@ export class StacheNavComponent implements OnInit, StacheNav {
 
     return (activeUrl === path);
   }
+
+  private isExternal(route: any): boolean {
+    let path = route.path;
+
+    if (typeof path !== 'string') {
```
Blackbaud-BrandonJones: Is there an instance where the path would not be a string?

src/app/public/src/modules/nav/nav.component.spec.ts
blackbaud-brandonhare: Whups! Fixing that.

src/app/public/src/modules/nav/nav.component.ts
blackbaud-brandonhare: Yes. Paths can be arrays too.

src/app/public/src/modules/nav/nav.component.spec.ts
```diff
@@ -101,6 +101,15 @@ describe('StacheNavComponent', () => {
     expect(route.isCurrent).toBe(true);
   });
 
+  it('should return true if a given route is external', () => {
+    component.routes = [{ name: 'Test', path: 'https://google.com' }];
+    const route = component.routes[0];
+    let result = component['isExternal'](route);
```
Blackbaud-SteveBrush: So, the reason coverage is failing is because you're calling the private method explicitly, instead of allowing it to be called implicitly via the public `navigate()` method. Just need to rework the unit test.

src/app/public/src/modules/nav/nav.component.spec.ts
```diff
@@ -101,6 +101,15 @@ describe('StacheNavComponent', () => {
...
+    fixture.detectChanges();
+
+    expect(result).toBe(true);
```
Blackbaud-SteveBrush: I take it back. This test is probably sufficient. See my comment below.

src/app/public/src/modules/nav/nav.component.spec.ts
```diff
@@ -121,6 +130,17 @@ describe('StacheNavComponent', () => {
...
 
+  it('should navigate to an external url', () => {
+    spyOn(component, 'isExternal');
+    component.routes = [{ name: 'Test', path: 'https://google.com' }];
+    const route = component.routes[0];
+
+    fixture.detectChanges();
+    component.navigate(route);
+
+    expect(component['isExternal']).toHaveBeenCalled();
```
Blackbaud-SteveBrush: So, I added a console log to the `if (this.isExternal())` block and it never gets called. I wonder if you could modify this test to spy on a mock window's `href` property, and check that instead?

PR #270, opened 2017-07-17
Feature action buttons search

blackbaud-brandonhare:
I'll work on tests next. Wanted some feedback on implementation first.


src/app/public/src/modules/action-buttons/action-buttons.component.ts
```diff
@@ -1,13 +1,42 @@
...
+  private originalRoutes: StacheNavLink[];
+
+  public ngOnInit() {
+    this.originalRoutes = this.routes.slice();
+  }
+
+  public searchApplied(searchText: string) {
+    let filteredItems = this.originalRoutes;
+    if (searchText) {
+      filteredItems = this.routes.filter(function (item: any){
```
Blackbaud-BrandonJones: `this.originalRoutes.filter`     Filtering against the routes causes an issue when you try to search a couple times, it continues to search against the filtered list. 
Search for 2 ->   search for 3 ->   list disappears completely.

blackbaud-brandonhare: Yeah, I know why. The `ngIf` check on there for routes existing. If you filter to no routes, it hides the component completely. I'll have to modify that check to account for that. Good catch.

src/app/public/src/modules/action-buttons/action-buttons.component.ts
```diff
@@ -1,13 +1,42 @@
...
+          if (item.hasOwnProperty(property) && (property === 'name' || property === 'summary')) {
+            if (item[property].toLowerCase().indexOf(searchText.toLowerCase()) > -1) {
+              return true;
+            }
+          }
+        }
+        return false;
+      });
+    }
+    this.routes = filteredItems;
```
Blackbaud-BrandonJones: We can simplify this,  don't change the this.routes, instead create a new `filteredRoutes`  we can iterate over in the `ngFor`, keeping the original routes in place.  This will also eliminate the need for a variable inside this , method.

```
  private filteredRoutes: StacheNavLink[];

  public ngOnInit() {
    this.filteredRoutes = this.routes.slice();
  }

  public searchApplied(searchText: string) {
    this.filteredRoutes = this.routes;
    if (searchText) {
      this.filteredRoutes = this.routes.filter(function (item: any){
        let property: any;
        for (property in item) {
          if (item.hasOwnProperty(property) && (property === 'name' || property === 'summary')) {
            if (item[property].toLowerCase().indexOf(searchText.toLowerCase()) > -1) {
              return true;
            }
          }
        }
        return false;
      });
    }
  }
```

src/app/public/src/modules/action-buttons/action-buttons.component.scss
```diff
@@ -9,6 +9,13 @@
   display: inline-flex;
 }
 
+.stache-action-buttons-search-container {
+  position: relative;
+  display: flex;
+  min-height: 49px;
+  margin-bottom: 10px;
+}
```
Blackbaud-BrandonJones: Some strange spacing because of the `stache-row` on the action buttons?   Do we want to mimic the `-15px` margins on the left and right that the `stache-row` implements?

Blackbaud-SteveBrush:
Hey, I know this started as a simple, copy-it-over-from-skyux, but I do have a few feature suggestions that would be useful.

If you captured the `searchText` as a public property on the component, you could do the search on a `keyup` event, instead of relying on a return or button click.

(I'm not married to the code below; it purely serves as an illustration.)

**The search field:**

```
<sky-search *ngIf="showSearch" expandMode="fit"
  [searchText]="searchText"
  (searchApply)="searchApplied($event)"
  (keyup)="onKeyup($event)">
</sky-search>
```

**onKeyup:**

```
// Capture the search field's value on keyup...
public onKeyup(event: KeyboardEvent) {
  const searchText = (event.target as HTMLInputElement).value;
  this.searchApplied(searchText);
}
```

**searchApplied:**

```
public searchApplied(searchText: string) {
  // Keep a list of searchable fields, to allow easy-edit in the future...
  const searchableFields = ['name', 'summary'];

  // Save the search text so the template can auto-populate it into the field.
  this.searchText = searchText;
  this.filteredRoutes = this.routes;
  const query = searchText.toLowerCase();

  if (!searchText) {
    return;
  }

  // Just a different way of looping through the action buttons:
  this.filteredRoutes = this.routes.filter((route: any) => {
    const matchingFields = searchableFields.filter((field: string) => {
      const isMatch = (route[field].toLowerCase().indexOf(query) > -1);
      return isMatch;
    });

    return (matchingFields.length > 0);
  });
}
```

Blackbaud-SteveBrush:
Great stuff! You're clear to adjust unit tests...

PR #276, opened 2017-07-19
Moved nav service to route service, created new nav service

Blackbaud-BrandonJones:
Depends on: https://github.com/blackbaud/skyux-builder/pull/232


src/app/public/src/modules/nav/nav.service.ts
```diff
@@ -1,160 +1,49 @@
...
-      }
-    });
-  }
+    private windowRef: StacheWindowRef) { }
 
-  public getActiveRoutes(): StacheNavLink[] {
-    if (this.activeRoutes) {
-      return this.activeRoutes;
+  public scrollToElement(fragment: string) {
+    let element = this.windowRef.nativeWindow.document.querySelector(`#${fragment}`);
```
blackbaud-brandonhare: Why querySelector over getElementById?

src/app/public/src/modules/nav/nav.service.ts
```diff
@@ -1,160 +1,61 @@
...
-        return metaRoute.path === route.path;
-      })[0];
-
-      if (foundRoute) {
-        return foundRoute;
+      // The current page is the path intended, but no element with the fragment exists, scroll to
+      // the top of the page and remove the hash from the url.
+      if (route.path === currentPath) {
+        this.windowRef.nativeWindow.scroll(0, 0);
+        this.windowRef.nativeWindow.location.hash = '';
```
Blackbaud-SteveBrush: Is there a reason we can't leave the hash in the URL? Seems it would be confusing for someone if the hash didn't work, and suddenly disappeared. If the hash remains and the window doesn't scroll, the user would just figure that the fragment is no longer valid.

Blackbaud-BrandonJones: We could leave it there for sure.   I just figure it would look cleaner without the hash.   Just a style decision, but fine with leaving it in there.

Blackbaud-SteveBrush: Second thought, this is fine.

Blackbaud-BrandonJones: 6 of 1 half dozen of the other.   It could be helpful if there is actually a broken hash someone tried to link for some reason to see where it went wrong.   Or it could just be distracting to have a hash that doesn't do anything but brings you to the top of the page.

src/app/public/src/modules/nav/nav.service.ts
```diff
@@ -29,7 +29,6 @@ export class StacheNavService {
       // the top of the page and remove the hash from the url.
```
Blackbaud-SteveBrush: "...and remove the hash from the url."

PR #288, opened 2017-08-01
Refactored Nav Order

Blackbaud-BrandonJones:
Refactored navOrder to allow for a single route to be moved around a nav list without the need for all items to be ordered.


src/app/public/src/modules/shared/route.service.ts
```diff
@@ -126,36 +126,55 @@ export class StacheRouteService {
...
 
-    const orderedRoutes = routes.filter(route => route.hasOwnProperty('order'))
-      .sort(sortByName)
-      .sort(sortByOrder);
+    return sortedRoutes;
+  }
 
-    const unorderedRoutes = routes.filter(route => !route.hasOwnProperty('order'))
-      .sort(sortByName);
+  private validateNavOrder = (route: any) => {
```
Blackbaud-SteveBrush: Mind formatting these methods in the "typescripty" way? Instead of using fat arrows...

src/app/public/src/modules/shared/route.service.ts
```diff
@@ -126,36 +126,55 @@ export class StacheRouteService {
...
-        return 0;
-      }
-    };
-
-    const sortByOrder = (a: any, b: any): number => {
-      if (a.order < b.order) {
-        return -1;
-      } else if (a.order > b.order) {
-        return 1;
+    const sortedRoutes = routes.map(this.validateNavOrder)
```
Blackbaud-SteveBrush: The route order is derived from the `StacheRouteMetadataService`, correct? I feel like the `order` should always be "valid" if it is fetched from the service. Maybe include the validation methods here?
https://github.com/blackbaud/stache2/blob/master/src/app/public/src/modules/shared/route-metadata.service.ts#L23

src/app/demos/route-metadata/route-order-test/index.html
```diff
@@ -0,0 +1,6 @@
+<app-nav></app-nav>
+<stache>
+  <stache-page-summary>
+    Route ordering... FIGHT!
```
blackbaud-brandonhare: TOO MUCH WHIMSY JONES!

Blackbaud-BrandonJones: I'll show you too much whimsy.

src/app/public/src/modules/shared/route-metadata.service.ts
```diff
@@ -19,11 +19,19 @@ export class StacheRouteMetadataService {
     @Inject(STACHE_ROUTE_METADATA_SERVICE_CONFIG)
     public metadata: any[]) {
       this.metadata.forEach(route => {
-        if (route.order) {
-          route.order = numberConverter(route.order);
+        if (route.hasOwnProperty('order')) {
+          route = this.validateNavOrder(route);
```
Blackbaud-SteveBrush: Throw the check for `order` into the validate method.
```
this.metadata.forEach((route: any) => this.validateNavOrder(route));
```

src/app/public/src/modules/shared/route-metadata.service.ts
```diff
@@ -19,11 +19,19 @@ export class StacheRouteMetadataService {
...
         }
       });
     }
+
+  private validateNavOrder(route: any): any {
+    route.order = numberConverter(route.order);
+    if (route.order <= 0 || route.order !== parseInt(route.order, 10)) {
```
Blackbaud-SteveBrush: What is `route.order !== parseInt(route.order, 10)` checking that `numberConverter` didn't already check?

Blackbaud-BrandonJones: This is making sure the value is not a decimal

src/app/public/src/modules/shared/route-metadata.service.ts
```diff
@@ -18,12 +18,17 @@ export class StacheRouteMetadataService {
...
-          route.order = numberConverter(route.order);
-        }
-      });
+      this.metadata.forEach((route: any) => this.validateNavOrder(route));
     }
+
+  private validateNavOrder(route: any): void {
+    if (route.hasOwnProperty('order')) {
+      route.order = numberConverter(route.order);
+      if (route.order <= 0 || route.order !== parseInt(route.order, 10)) {
```
Blackbaud-SteveBrush: I learned something new today! Would you mind putting the logic in a variable for the sake of readability?

Also, aren't we supporting zero? I think we should.

```
const isDecimal = (route.order !== parseInt(route.order, 10));
if (route.order < 0 || isDecimal) {
  delete route.order;
}
```

src/app/public/src/modules/shared/route.service.ts
```diff
@@ -126,36 +126,45 @@ export class StacheRouteService {
...
-      .sort(sortByOrder);
+    });
 
-    const unorderedRoutes = routes.filter(route => !route.hasOwnProperty('order'))
-      .sort(sortByName);
+    return sortedRoutes;
+  }
 
-    const sortedRoutes = orderedRoutes.concat(unorderedRoutes);
+  private sortByName(a: any, b: any): number {
```
Blackbaud-SteveBrush: 👍

src/app/public/src/modules/shared/route-metadata.service.ts
Blackbaud-BrandonJones: I thought about this, and decided against supporting 0.  It created the same problem supporting `-` numbers did.   If I had a `navOrder='0'` somewhere in a large list.   Later I wanted to move a nav item around to the 4th position in a list,   I would immediately try `navOrder='4'`... and then be surprised when it went to the 5th position.     

By not supporting `0`, this allows our `navOrder` to directly reflect it's position in the list.   (unless there are duplicate navOrder items).

src/app/public/src/modules/shared/route.service.spec.ts
```diff
@@ -206,14 +214,29 @@ describe('StacheRouteService', () => {
...
     expect(activeRoutes[0].children[0].children[2].name).toBe('A Three');
     expect(activeRoutes[0].children[0].children[2].order).toBe(3);
     expect(activeRoutes[0].children[0].children[3].name).toBe('B Three');
     expect(activeRoutes[0].children[0].children[3].order).toBe(3);
-    expect(activeRoutes[0].children[0].children[6].name).toBe('A');
   });
+
+  it('should order routes to the order index and leave non ordered routes before it',
+    () => {
+          router.url = '/order-routes';
```
Blackbaud-SteveBrush: You got 4 spaces here, instead of 2 :)

src/app/demos/route-metadata/route-order-test/j/index.html
```diff
@@ -0,0 +1,5 @@
+<stache
+  pageTitle="j"
+  navOrder="0">
+   0
```
Blackbaud-SteveBrush: This page doesn't display at the top of the list.

src/app/public/src/modules/shared/route.service.ts
```diff
@@ -126,36 +126,45 @@ export class StacheRouteService {
...
+      .sort(this.sortByName);
+
+    const routesWithNavOrder = routes.filter(route => route.hasOwnProperty('order'))
+      .sort(this.sortByName)
+      .sort(this.sortByOrder);
+
+    routesWithNavOrder.forEach(route => {
+      let newIdx = route.order - 1;
+      if (newIdx < sortedRoutes.length) {
+        while (newIdx < sortedRoutes.length && sortedRoutes[newIdx].order <= route.order) {
```
Blackbaud-SteveBrush: I'm sure this logic is sound, but I had a difficult time dissecting it. If there's any way to simplify with inline methods or wrapping expressions in a single boolean variable, it would help my aging brain. Love the demo you created for this! Helps visualize exactly what you're going for.

src/app/demos/route-metadata/route-order-test/j/index.html
Blackbaud-BrandonJones: Correct,  we are not accepting 0 as a valid order.   In this case we are not 0 indexing our 'ordering'.   Our client base would expect to put a `1`  and it be the first item in the nav and a `4` to be the 4th, not off by one as would occur with a 0 indexed version.

src/app/public/src/modules/shared/route-metadata.service.ts
blackbaud-brandonhare: Yeah, I'm in favor of not supporting 0 in this case too. It makes more sense for what it's doing on the page, and zero based indexes aren't as intuitive to non-technical users.

blackbaud-brandonhare: Documentation can cover for any potential confusion, but I think having ordering from 1 to whatever makes sense.

PR #297, opened 2017-08-28
Feature blockquote

Blackbaud-BrandonJones:
screen shot 2017-08-29 at 3 32 12 pm

Blockquote component.


Blackbaud-SteveBrush:
This looks really good!

I believe it needs the `$stache-line-height` applied (throw some lorem ipsum in there to double-check, http://www.lipsum.com/feed/html).

Also, can you see what the light grey color looks like as the background? Just to give it a bit more visual difference to the table of contents.

src/app/public/src/modules/blockquote/blockquote.component.html
```diff
@@ -0,0 +1,15 @@
+<div class="stache-blockquote">
+  <div class="stache-blockquote-icon">
+    <i class="fa fa-quote-left" aria-hidden="true"></i>
```
blackbaud-brandonhare: Some accessibility? Nice! We should add a task to do an accessibility assessment for our stuff. Maybe Matt Gregg could help us with that.

src/app/public/src/modules/blockquote/blockquote.component.scss
```diff
@@ -0,0 +1,27 @@
+@import '../../styles/_variables';
+
+.stache-blockquote {
+  border-left: 4px solid $stache-background-color-primary-dark;
+  font-size: $stache-font-size-base * 1.12;
+  color: $stache-text-color-deemphasized;
+  padding: $stache-padding * 3;
```
Blackbaud-SteveBrush: This PR is perfect except for the padding. The padding-left doesn't match up with the table of contents. 

Change this to `15px 20px` and merge!

PR #298, opened 2017-08-31
fixed a bug not allowing us to get the appname if none was set in sky…

Blackbaud-BrandonJones:
…ux config


src/app/public/src/modules/analytics/google-analytics.directive.ts
```diff
@@ -81,6 +83,8 @@ export class StacheGoogleAnalyticsDirective implements OnInit {
       this.analyticsClientId = appSettings.stache.googleAnalytics.clientId;
     }
 
+    this.appName = this.configService.runtime.app.base.replace(/^\/|\/$/g, '');
```
Blackbaud-SteveBrush: Could `runtime.app.base` ever be empty? If so, we should probably check for that and set `isEnabled` to false. 

If it is _never_ empty, this looks good to me!

Blackbaud-BrandonJones: It can't be if you ever serve your app.   That setting is what builder uses to serve the app.

https://github.com/blackbaud/skyux-builder/blob/0378241743ad5d06408127e52513bf8a0ebcd3db/config/sky-pages/sky-pages.config.js#L98

PR #306, opened 2017-09-11
Added swagger ui component to stache.

blackbaud-brandonhare:


src/app/public/src/modules/swagger/swagger.component.ts
```diff
@@ -0,0 +1,33 @@
...
+
+@Component({
+  selector: 'stache-swagger',
+  templateUrl: './swagger.component.html'
+})
+export class StacheSwaggerComponent implements OnInit, AfterViewInit {
+  @Input()
+  public swaggerUrl: string;
+
+  private swaggerConfig = {
```
Blackbaud-BrandonJones: Is this something that can live in the skyuxConfig similar to the analytics stuff?
https://github.com/blackbaud/stache2/blob/master/skyuxconfig.json#L13

Blackbaud-BrandonJones:
Wasn't Swagger officially being changed to be called `Open API` or something like that?  Should we update these component names to use that instead?

Blackbaud-LindseyRix:
Yes, the whole thing is now called Open API. But, I think there's a piece that's still called Swagger.

src/app/public/src/modules/swagger/swagger.component.ts
blackbaud-brandonhare: I wasn't sure if this was something we wanted to override, but yeah, I can move it to the config. I was initially going to put it there anyway.

blackbaud-brandonhare:
This portion is still called Swagger UI.

src/app/public/src/modules/swagger/swagger.component.ts
blackbaud-brandonhare: And we can't pull all of it out regardless. Since some of that stuff isn't static.

src/app/public/src/modules/swagger/swagger.component.html
```diff
@@ -0,0 +1 @@
+<div id="swagger-ui"></div>
```
Blackbaud-SteveBrush: Would it be possible to replace this with a class name, and keep convention with the other components? This will keep the feature modular, and not tie a component instance to a specific identifier.

```
<div class="stache-swagger-ui" #swaggerWrapper></div>

// ...

swagger.SwaggerUIBundle({
  // dom_id: '#swagger-ui',
  domNode: this.swaggerWrapper
});
```

blackbaud-brandonhare:
It is not possible. It’s a requirement of the swagger-ui api.

src/app/public/src/modules/swagger/swagger.component.html
blackbaud-brandonhare: It's not possible, since it's a requirement of the api, but I can change it to #stache-swagger-ui.

Blackbaud-SteveBrush: @blackbaud-brandonhare : Did you see my inclusion of `domNode`? https://github.com/swagger-api/swagger-ui#parameters

Blackbaud-SteveBrush: ^ I'm setting `domNode` with a `ViewChild` reference...

src/app/public/src/modules/swagger/swagger.component.ts
blackbaud-brandonhare: Yeah, it's not worth pulling this out. We would literally only have the default URL and the template in here.

src/app/public/src/modules/swagger/swagger.component.ts
```diff
@@ -0,0 +1,33 @@
...
+    // tslint:disable-next-line:max-line-length
+    url: 'https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/examples/v2.0/json/petstore-expanded.json',
```
Blackbaud-SteveBrush: I may be reading this incorrectly, but there doesn't seem to be a reason to provide this value (`url`), since it will be overwritten in `ngOnInit` (even if it's undefined)?

PR #310, opened 2017-09-11
Added alt tag to stache image.

blackbaud-brandonhare:


src/app/home.component.html
```diff
@@ -87,6 +87,7 @@
   </stache-page-anchor>
 
   <stache-image
+    alt="This is an image"
```
Blackbaud-SteveBrush: Screen readers might have issue with this (`alt` isn't just for images). http://webaim.org/techniques/alttext/#conclusion

I'd prefer `imageAlt` or `alternateText`, instead.

src/app/public/src/modules/image/image.component.ts
```diff
@@ -8,4 +8,7 @@ import { Component, Input } from '@angular/core';
 export class StacheImageComponent {
   @Input()
   public imageSource: string;
+
+  @Input()
+  public imageAlt: string;
```
Blackbaud-SteveBrush: Set the default to `''` so that we don't get:

screen shot 2017-09-12 at 10 56 29 am

blackbaud-brandonhare: Ah, good call.

PR #317, opened 2017-09-18
Omnibar sensitivity 

Blackbaud-BrandonJones:
fix for the sidebar nav in affix, style fix for page anchors to account for omnibar.

A longer term solution may be needed to figure out a means of manually writing a `scrollIntoView` method so we can put in an `offset` value to adjust the scroll height for when the omnibar exists.


src/app/public/src/modules/page-anchor/page-anchor.component.scss
```diff
@@ -26,3 +26,8 @@
   height: 0.65em;
   transition: opacity 300ms;
 }
+
+.anchor-with-omnibar {
+  padding-top: 5em;
+  margin-top: -3em;
```
Blackbaud-SteveBrush: This is brilliant!

I'm not sure you need the `.anchor-with-omnibar` class, though. You can just put this directly on the `.stache-page-anchor` class and save yourself the component conditional.

Blackbaud-SteveBrush: Also, I'd double-check that this behavior works the same way on Windows browsers :)

Blackbaud-BrandonJones: I added the class because without the omnibar it scrolled to the anchor, but a bit high, around 45px higher than seemed normal.

Blackbaud-SteveBrush:
I'm fairly certain you could avoid almost all of the new logic in the Affix Top directive. All you really need to check is the height of the omnibar, and adjust the `windowIsScrolledBeyondElement` boolean accordingly:

**affix-top.directive.ts:**

```
const omnibarHeight = this.omnibarAdapterService.getHeight();
const windowIsScrolledBeyondElement =
  (this.offsetTop - omnibarHeight <= this.windowRef.nativeWindow.scrollY);
```

Then, everything else can be governed by styles:

**Add to layout-sidebar.component.scss:**
```
/deep/
.stache-has-omnibar { // added to the body tag
  .stache-affix-top {
    @media (min-width: $stache-screen-sm-min) {
      margin-top: 50px;

      .stache-sidebar {
        padding-bottom: 50px;
      }
    }
  }
}
```

The `.stache-has-omnibar` class would preferably be applied to the `body` tag, by an "adapter" service. I can imagine other components needing this class name in the future (perhaps the Help widget?).

SKY UX uses adapter services to interact with the DOM. In this case, the adapter service would be interacting with the Omnibar's iframe, and, 1) checking if it exists, 2) adding classnames to the body, and 3) returning the Omnibar's height.

**omnibar-adapter.service.ts:**

```
  private element = document.querySelector('.sky-omnibar-iframe') as HTMLElement;
  public getHeight(): number { // ... }
  private applyClassName() { // ... }
}
```

My reasoning for all this is that this feature is primarily _visual_, and the logic should be pulled out of the components if at all possible. I'm willing to talk this through if needed!

src/app/public/src/modules/page-anchor/page-anchor.component.scss
Blackbaud-BrandonJones: @Blackbaud-SteveBrush confirmed, it does work in IE

Blackbaud-SteveBrush: Another idea (and this will also allow you to remove the omnibar logic in this component), you could use the `body` class in the stylesheet:

```
.stache-page-anchor {
  padding-top: 2em;
  margin: 0 0 1em;
  padding-bottom: 0.6em;
  border-bottom: 1px solid $stache-border-color-neutral-light;

  &:hover {
    .stache-page-anchor-icon {
      opacity: 1;
    }
  }

  /deep/
  .stache-omnibar-enabled & {
    padding-top: 5em !important;
  }
}
```

The above will at least get you to feature-parity, without the extra logic in the component. However, I'm curious about something else: It would be ideal if the top padding/margin of the Page Anchors was consistent whether or not the Omnibar was present. I know you mentioned the smooth-scroll method earlier. Did you discover a potential solution with that?

Blackbaud-BrandonJones: The solution with the smooth-scroll stuff was to adjust the padding and go negative margin.  The scroll goes to the top of the element (padding, not margin).  There is no way to set an offset of pixels or the like.  If we kept the padding/margin the same with or without omnibar, the element specified would just show up farther down the page than it would appear with the omni bar enabled.   If that's acceptable we can just always do the positive padding negative margin.

Blackbaud-SteveBrush: Thanks for the explanation. I think if you can use the `body` class, along with the positive/negative padding/margin, this would work great! (Updated example above.)

src/app/public/src/modules/shared/omnibar-adapter.service.ts
```diff
@@ -0,0 +1,33 @@
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class StacheOmnibarAdapterService {
+  private static readonly HAS_OMNIBAR_CLASS_NAME: string = 'stache-omnibar-enabled';
```
Blackbaud-SteveBrush: Love the readonly, here :)

src/app/public/src/modules/shared/omnibar-adapter.service.ts
```diff
@@ -0,0 +1,33 @@
...
+  private element: any = document.querySelector('.sky-omnibar-iframe');
```
Blackbaud-SteveBrush: I know it's a pain, but any use of the DOM should be wrapped in the `StacheWindowRefService` :-/

(Same as for `applyClassToBody` below.)

PR #320, opened 2017-09-22
[HOLD] - Feature search service

blackbaud-brandonhare:


src/app/public/src/modules/shared/stache-search-results-provider.ts
```diff
@@ -0,0 +1,130 @@
...
+import { StacheConfigService } from './config.service';
+import 'rxjs/add/operator/toPromise';
+
+@Injectable()
+export class StacheSearchResultsProvider {
+
+  private static readonly STACHE_QUERY_URL:
+  string = 'https://stache-search-query.sky.blackbaud.com';
+  private static readonly STACHE_QUERY_ENDPOINTS:
+  any = {
```
Blackbaud-SteveBrush: Possible to put the `string` and `any` on the same line as the variable name?

src/app/public/src/modules/shared/stache-search-results-provider.ts
```diff
@@ -0,0 +1,130 @@
...
+    private http: SkyAuthHttp,
+    private config: StacheConfigService
+  ) {
+    this.initIsInternal(this.config);
+    this.initEndpoints(this.config);
+    this.initSiteNames(this.config);
+    this.initUrl(this.config);
+  }
+
+  public getSearchResults(searchArgs: any) {
```
Blackbaud-SteveBrush: Add a return type, please.

src/app/public/src/modules/shared/stache-search-results-provider.ts
```diff
@@ -0,0 +1,130 @@
...
+          };
+        }
+        let results = {
+          items: response.results.map((item: any) => {
+            return {
+              title: item.document.title,
+              subtitle: item.document.site_name,
+              description: this.parseDescription(item),
+              url: item.document.host + '/' + item.document.site_name + item.document.path
+            };
```
Blackbaud-SteveBrush: It would benefit consuming classes/apps to provide an interface for the results.

src/app/public/src/modules/shared/stache-search-results-provider.ts
```diff
@@ -0,0 +1,130 @@
...
+          })
+        };
+        return results;
+      })
+      .catch(err => console.log(err));
+  }
+
+  private getQueryUrl() {
```
Blackbaud-SteveBrush: Return type.

src/app/public/src/modules/shared/stache-search-results-provider.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  private initUrl(config: any) {
+    if (this.checkConfig(config, 'url')) {
+        this.url = this.config.skyux.appSettings.stache.searchConfig.url;
+      } else {
+        this.url = StacheSearchResultsProvider.STACHE_QUERY_URL;
+      }
+  }
+
+  private checkConfig(config: any, key?: string) {
```
Blackbaud-SteveBrush: Return type.

src/app/public/src/modules/shared/stache-search-results-provider.ts
```diff
@@ -0,0 +1,130 @@
...
+    ) {
+      if (key) {
+        return (config.skyux.appSettings.stache.searchConfig[key] !== undefined) ? true : false;
+      }
+      return true;
+    }
+    return false;
+  }
+
+  private parseDescription(item: any) {
```
Blackbaud-SteveBrush: Return type.

Blackbaud-SteveBrush: Also, it would benefit the IDE to provide an interface for `item`.

Blackbaud-SteveBrush:
Mind removing `stache-` from the provider's file name? All the other files honor this convention...

src/app/public/src/modules/shared/stache-search-results-provider.ts
blackbaud-brandonhare: I'm the consuming class/app in this case. I'm conforming to what the Omnibar search provider is expecting.

blackbaud-brandonhare: I'm sure an interface will get added for that at some point.

src/app/public/src/modules/shared/search-result.ts
```diff
@@ -0,0 +1,13 @@
+export interface SearchResult {
```
Blackbaud-SteveBrush: `Stache` prefix, please.

src/app/public/src/modules/shared/search-results-provider.ts
```diff
@@ -0,0 +1,129 @@
...
+    private http: SkyAuthHttp,
+    private config: StacheConfigService
+  ) {
+    this.initIsInternal(this.config);
+    this.initEndpoints(this.config);
+    this.initSiteNames(this.config);
+    this.initUrl(this.config);
+  }
+
+  public getSearchResults(searchArgs: any): Promise<any> {
```
Blackbaud-SteveBrush: Does this return a `Promise<StacheSearchResult[]>`?

blackbaud-brandonhare: No. The stache search result is what comes from my service. The return from that is the interface that skyux established for the search provider.

Blackbaud-SteveBrush: Understood!

blackbaud-brandonhare:
Probably release them all at once as soon as I get my services deployed to prod!

src/app/public/src/modules/shared/search-results-provider.ts
```diff
@@ -0,0 +1,143 @@
...
+  private isInternal: boolean = true;
+  private siteNames: string[] = [];
+  private endpoints: any = {
+    internal: 'query',
+    public: 'query-public'
+  };
+  private queryUrlBase: string = 'https://stache-search-query.sky.blackbaud.com';
+
+  constructor(
+    private http: SkyAuthHttp,
```
Blackbaud-BrandonJones: Can you do a conditional setting of http?  Try to set the optional `SkyAuthHttp`, then in the constructor, if http does not exist, set it to a new instance of Http from angular?
`@Optional() http = SkyAuthHttp`  
```
if (!http) {
  this.http = new Http();
}
```

package.json
```diff
@@ -35,13 +35,15 @@
     "@blackbaud/skyux": "^2.0.0"
   },
   "dependencies": {
-    "@blackbaud/skyux-builder-plugin-stache": "^1.0.0",
+    "@blackbaud/skyux-builder-plugin-stache": "https://github.com/blackbaud/skyux-builder-plugin-stache.git#feature-search-service",
```
Blackbaud-SteveBrush: ?

src/app/public/src/modules/shared/search-results-provider.spec.ts
```diff
@@ -0,0 +1,391 @@
+import { expect } from '@blackbaud/skyux-builder/runtime/testing/browser';
+import { Response, ResponseOptions } from '@angular/http';
+import { Observable } from 'rxjs';
+
+import { StacheSearchResultsProvider, StacheConfigService } from './index';
+
+class MockHttpService {
+  public post = jasmine.createSpy('post').and.callFake((url: string, body: any) => {
+    console.log(url);
+    console.log(body);
```
Blackbaud-SteveBrush: Logs needed?

blackbaud-brandonhare: Yes. I'm spying on those to check values.

src/app/public/src/modules/shared/search-results-provider.ts
```diff
@@ -0,0 +1,151 @@
...
+  private siteNames: string[] = [];
+  private endpoints: any = {
+    internal: 'query',
+    public: 'query-public'
+  };
+  private queryUrlBase: string = 'https://stache-search-query.sky.blackbaud.com';
+
+  constructor(
+    @Optional() private authHttp: SkyAuthHttp,
+    private regHttp: HttpClient,
```
Blackbaud-BrandonJones: Does this have to be injected like this?  We can't conditionally create a new one in the constructor itself?

blackbaud-brandonhare: Not unless we want to do conditional injection, which is an anti-pattern.

Blackbaud-BrandonJones: https://stackoverflow.com/questions/43450259/angular-2-conditionally-inject-service-in-component
? The Injector pattern?

blackbaud-brandonhare: Yeah, I saw that post too. If you read further down, they explicitly say to avoid it. I think this solution circumvents that issue and is much more readable. I could add a comment for clarity if you think that would help. Thoughts @Blackbaud-SteveBrush?

Blackbaud-BrandonJones: Just trying to see if there is an option we can use to avoid having an un-used injected class if the auth exists.
```
constructor(
@Optional() private http: SkyAuthHttp
) {
 if(!this.http) {
   this.http = <Http>this.injector.get(Http);
}
}
```

blackbaud-brandonhare: If it's that big of a concern (and I'm not convinced it is), then really we should create a wrapper service for it and inject that instead rather than using the injector directly.

Blackbaud-SteveBrush: I agree that this is messy. Ideally, you would use a [provider factory](https://stackoverflow.com/questions/40027892/angular-cli-conditional-imports-using-an-environment-variable). If this proves too difficult, I would at the very least isolate this logic into a separate service (such as `StacheSearchResultsHttp`) that can be injected into this provider's constructor. The http service would then determine which client is should use.

Also, curious why you added the suffix "Provider" to this class, instead of "Service"?

blackbaud-brandonhare: I think the factory will be fine. We can abstract it there. Not sure how testing that will be though. Kinda seems like it will be terrible. And I'm using Provider to be in line with the SKYUX and Omnibar naming conventions for that particular thing. That's what Paul names them.

PR #342, opened 2017-11-02
Feature auth http service

blackbaud-brandonhare:
This is the only method that works. Requires the builder release and the new http-auth stache plugin update.


src/app/public/src/modules/shared/http.service.ts
```diff
@@ -0,0 +1,7 @@
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class StacheHttpService {
```
Blackbaud-BobbyEarl: Should this class be `abstract`?

blackbaud-brandonhare: The config service one is not abstract, so I'm following the pattern there. @Blackbaud-SteveBrush may have more insight on that one.

PR #364, opened 2017-11-30
Updated version

blackbaud-brandonhare:


Blackbaud-BrandonJones:
I feel like this is an unecessary version bump

blackbaud-brandonhare:
We absolutely want to set the version requirement to 1.3 for the latest. Otherwise the search pipeline stuff will not work.

Blackbaud-BrandonJones:
Yeah, but `^1.2.0` === `^1.3.1`,  the carrot the start of the version number will use the lastest minor release.

blackbaud-brandonhare:
Right, but we want to enforce 1.3 if possible. I'm not saying we should do a new release or anything. I can set it to `^1.3` and be done with it.

Issue #368, opened 2017-12-04
Page title overlaps breadcrumb after scrolling


Blackbaud-TommyVernieri:
### Expected behavior

Expected sidebar title to be in line with the top of the main page body.

### Actual behavior

Page title overlaps with breadcrumbs

### Steps to reproduce

1. Navigate to https://developer.blackbaud.com/stache/learn/get-started
2. Scroll down so that the header and breadcrumbs leave the viewport
3. Scroll up to the top of the page
4. Notice that the title of the page in the sidebar overlaps with the breadcrumbs

Happens intermittently. Reproduced sometimes on Firefox 57, but not every time.

![image](https://user-images.githubusercontent.com/13053902/33568532-b898a750-d8f4-11e7-8d6f-5cedddebd1df.png)

PR #374, opened 2017-12-04
affix overlap fix

Blackbaud-BrandonJones:
Added a method to recalculate omnibar height if it's larger than anticipated sizes.

There was an issue where loading the new page with the omnibar from a dropdown was causing the omnibar height to be calculated as the full screen because of how the iframe works with drop-downs with the new omnibar.  This method re-call itself while the omnibar remains too large when loading a new page.

fixes: https://github.com/blackbaud/stache2/issues/368


src/app/public/src/modules/shared/omnibar-adapter.service.ts
```diff
@@ -25,7 +27,24 @@ export class StacheOmnibarAdapterService {
   }
 
   private setHeight(): void {
-    this.omnibarHeight = this.element.offsetHeight;
+    let omnibarElementHeight = this.element.offsetHeight;
+
+    if (this.omnibarIsExpanded(omnibarElementHeight)) {
+
+      if (this.setHeightAttempts < 10) {
```
blackbaud-brandonhare: Nice! Are we going to run this check for 5 seconds? Is that too long? I'm not sure personally, but that feels like a long time to wait for a load. But maybe it's better to be safe. It's not like this is an expensive operation.

Blackbaud-BrandonJones: It' won't delay loading. All this will do is give it a max of 5 seconds for the omnibar to finish collapsing back to it's standard size before saying "forget it"  and moving on.

blackbaud-brandonhare: Right, I just meant it was a long time to wait for the omnibar to load. But honestly, it probably won't be much of an issue.

src/app/public/src/modules/shared/omnibar-adapter.service.ts
```diff
@@ -5,8 +5,10 @@ import { StacheWindowRef } from './window-ref';
 @Injectable()
 export class StacheOmnibarAdapterService {
   private static readonly HAS_OMNIBAR_CLASS_NAME: string = 'stache-omnibar-enabled';
+  private static readonly EXPECTED_OMNIBAR_MAX_HEIGHT: number = 100;
```
Blackbaud-SteveBrush: Why `100`, instead of `50`?

Blackbaud-BrandonJones: I wanted a bit of a buffer just in case for some reason the `Omnibar` changes height by a few pixels.  Otherwise we could just not bother detecting the bar and just entering the static height.

src/app/public/src/modules/shared/omnibar-adapter.service.ts
```diff
@@ -5,29 +5,27 @@ import { StacheWindowRef } from './window-ref';
...
   public getHeight(): number {
-    return this.omnibarHeight;
+    if (this.omnibarEnabled()) {
+      return StacheOmnibarAdapterService.EXPECTED_OMNIBAR_HEIGHT;
+    }
+    return 0;
   }
 
   public omnibarEnabled(): boolean {
     // Converts the elements existence to a boolean.
```
blackbaud-brandonhare: element's

PR #378, opened 2017-12-11
removing smooth scroll

Blackbaud-BrandonJones:
Smooth scroll was buggy and inconsistent,  removing it for now so the functionality is there and consistent across browsers.


Blackbaud-BobbyEarl:
The `stache-page-anchor` component still uses `scrollIntoView`.

https://github.com/blackbaud/stache2/blob/e5be8f4d0f424b3e21b69a97d4b8e4a87888e326/src/app/public/src/modules/page-anchor/page-anchor.component.ts#L45

Issue #360, opened 2017-11-27
Investigate direct link to git file.


Blackbaud-BrandonJones:
Investigate adding a direct link mechanism to link to the `git file` for specific pages.  Requested by @Blackbaud-TimothyWittig "It makes it a little harder to contribute or view the history of a file with out that link".  This may be a one off for Engineering System.

Blackbaud-LindseyRix:
This is the old feature we had in Stache 1 that enables you to click the edit button from a page. Would apply to internal sites and public sites enabling community contribution model only.

Blackbaud-TommyVernieri:
The tutorial in blackbaud/stache2-docs#90 can help with this.

Blackbaud-LindseyRix:
Edit button feature released. https://github.com/blackbaud/stache2/pull/379

PR #379, opened 2017-12-14
Feature edit button

blackbaud-brandonhare:
~~~Depends on https://github.com/blackbaud/skyux-builder-plugin-stache/pull/34 getting merged first. You'll need to pull in this branch for the plugin to test.~~~

Addresses: https://github.com/blackbaud/stache2/issues/360


src/app/public/src/modules/edit-button/edit-button.component.ts
```diff
@@ -0,0 +1,59 @@
...
+import { StacheJsonDataService } from '../shared';
+import { StacheNavService } from '../nav';
+
+@Component({
+  selector: 'stache-edit-button',
+  templateUrl: './edit-button.component.html',
+  styleUrls: ['./edit-button.component.scss']
+})
+export class StacheEditButtonComponent implements OnInit {
+  @Input()
```
Blackbaud-BrandonJones: When would we ever want to change this text to something other than 'Edit' ?  As this is a simple button component that builds it's own url off the config, I don't see this being manually added anywhere that would need alterable text.

src/app/public/src/modules/edit-button/edit-button.component.ts
```diff
@@ -0,0 +1,59 @@
...
+  private routeFragment: string;
+
+  public constructor (
+    private dataService: StacheJsonDataService,
+    private navService: StacheNavService,
+    private router: Router
+  ) { }
+
+  public ngOnInit() {
+    this.routeFragment = this.router.url.split('#')[0].substring(1);
```
Blackbaud-BrandonJones: if you pull in the `StacheRouteService` you can make use of the method we already have written
https://github.com/blackbaud/stache2/blob/master/src/app/public/src/modules/shared/route.service.ts#L53

src/app/public/src/modules/edit-button/edit-button.component.ts
```diff
@@ -0,0 +1,59 @@
...
+  @Input()
+  public editButtonText = 'Edit';
+
+  private url: string;
+  private routeFragment: string;
+
+  public constructor (
+    private dataService: StacheJsonDataService,
+    private navService: StacheNavService,
+    private router: Router
```
Blackbaud-BrandonJones: Any reason we are pulling in the router instead of the StacheRouteService?
https://github.com/blackbaud/stache2/blob/master/src/app/public/src/modules/shared/route.service.ts#L9

src/app/public/src/modules/wrapper/wrapper.component.ts
```diff
@@ -74,6 +78,18 @@ export class StacheWrapperComponent implements OnInit, AfterContentInit, OnDestr
     this.destroyPageAnchorSubscriptions();
   }
 
+  private checkShowEditSetting(): boolean {
+    if (
+      this.config.skyux.appSettings
+      && this.config.skyux.appSettings.stache
+      && this.config.skyux.appSettings.stache.showEdit !== undefined
```
Blackbaud-BrandonJones: `return this.config.skyux.appSettings.stache.showEdit || false;`

src/app/public/src/modules/edit-button/edit-button.component.ts
Blackbaud-LindseyRix: I could see public-facing SPAs being particular about what language they use for the Edit button. I think we can enforce it internally, but it's naive to think no one will ever want to customize the name of the button.

Blackbaud-BrandonJones: I think we should be consistent with the language we use for this component across all SPAs that want to make use of it. With that being an input, in order for consumers to make use of this button, they wouldn't be able to globally turn it on.  They would have to manually add the button with that input name on *every page* they wanted it to display.  

With how this component takes care of all the other stuff for you (formatting the URLs etc.) Not being prescriptive and deliberate with the language we use seems out of place, especially with how difficult it would be to circumnavigate.

I would err on the side of maybe changing the text from `Edit` to `Contribute` or something possibly, but I would not leave it up to individual SPAs to determine their best word usage here.

blackbaud-brandonhare: I could see making this a global config setting as well.

src/app/public/src/modules/edit-button/edit-button.component.ts
blackbaud-brandonhare: I didn't know that existed. :P That's why. I'll use that instead.

src/app/public/src/modules/edit-button/edit-button.component.html
```diff
@@ -0,0 +1,7 @@
+<button
```
Blackbaud-SteveBrush: Mind wrapping this in a `<div class="stache-edit-button">` to make things easier to style in the future (or from another component)? If you don't like the DIV, adding a special class to the button would be fine, too: `stache-btn-edit`.

src/app/public/src/modules/edit-button/edit-button.component.scss
```diff
@@ -0,0 +1,3 @@
+button {
+  float: right;
```
Blackbaud-SteveBrush: This style would probably be better to put in the wrapper component (`.stache-wrapper ::ng-deep .stache-btn-edit {}`). Otherwise, wherever we put the edit button, it will always float right (if we wanted to put the button somewhere else for example).

blackbaud-brandonhare: Hmm, good point.

src/app/public/src/modules/edit-button/edit-button.component.ts
Blackbaud-SteveBrush: I can see the point of maintaining consistency between SPAs, but I don't think it's a huge deal to leave it up to the consumer to make this say whatever they want, "Edit on GitHub", "Edit on VSTS", etc.

src/app/public/src/modules/edit-button/edit-button.component.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  public onEditButtonClick() {
+    this.navService.navigate({ path: this.url });
+  }
+
+  private getUrl(): string {
+    const repoData = this.dataService.getByName('repo');
+    let frag = encodeURIComponent(this.routeService.getActiveUrl() + '/');
+    let result = repoData.url + frag + 'index.html';
```
Blackbaud-SteveBrush: Non-blocking, but `frag` and `result` can be `const`, here, since they are never changed.

src/app/public/src/modules/edit-button/edit-button.component.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    return repoData.vstsBranchSelector === undefined ? result : result + repoData.vstsBranchSelector;
```
Blackbaud-SteveBrush: Again, non-blocking style comment: parenthesis are a good indicator that an expression should be evaluated as boolean: `(repoData.vstsBranchSelector === undefined)`.

src/app/public/src/modules/wrapper/wrapper.component.ts
```diff
@@ -74,6 +78,15 @@ export class StacheWrapperComponent implements OnInit, AfterContentInit, OnDestr
     this.destroyPageAnchorSubscriptions();
   }
 
+  private checkShowEditSetting(): boolean {
+    if (
+      this.config.skyux.appSettings
+      && this.config.skyux.appSettings.stache
+    ) {
+      return this.config.skyux.appSettings.stache.showEdit || false;
+    }
```
Blackbaud-SteveBrush: There's a chance that this method won't return a boolean. Might need to rework the logic so that every branch returns true or false.

src/app/public/src/modules/layout/layout-sidebar.component.html
```diff
@@ -4,6 +4,9 @@
     <stache-breadcrumbs *ngIf="showBreadcrumbs" [routes]="breadcrumbsRoutes">
     </stache-breadcrumbs>
 
+    <stache-edit-button *ngIf="showEditButton">
+    </stache-edit-button>
```
Blackbaud-SteveBrush: Here's an untested idea, that might allow the user to provide their own `<stache-edit-button>` in the template:

```
<stache-edit-button *ngIf="showEditButton">
</stache-edit-button>
<ng-content
  *ngIf="!showEditButton"
  select="stache-edit-button">
 </ng-content>
```
You could simply add the above to the layouts HTML. If the user sets `showEditButton` on the wrapper, it uses the defaults, otherwise it will attempt to pull the component itself into the template.

Blackbaud-SteveBrush: The trick with this one is the check for `!showEditButton`. It might be confusing if the user types `showEditButton="false"`, and then expects their custom button to show up. Maybe a check for "undefined" would be better...

src/app/public/src/modules/wrapper/wrapper.component.ts
blackbaud-brandonhare: This will always return a boolean. If showEdit is undefined, which is a falsey value, it'll return false, which is the desired behavior. I could add another check and handle the undefined explicitly, but I don't see a need to.

blackbaud-brandonhare: I have a better way to handle this now, so this won't be an issue any more.

Blackbaud-BrandonJones: :clinking_glasses:

src/app/public/src/modules/wrapper/wrapper.component.ts
```diff
@@ -8,9 +8,11 @@ import { Subscription } from 'rxjs';
 
 import { StacheTitleService } from './title.service';
 import { StachePageAnchorComponent } from '../page-anchor';
-import { StacheJsonDataService } from '../shared';
+import { StacheConfigService, StacheJsonDataService } from '../shared';
 import { StacheNavLink, StacheNavService } from '../nav';
 
+const get = require('lodash.get');
```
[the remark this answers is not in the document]
Blackbaud-BrandonJones: Agreed with a separate PR for making this whole thing a method on the config service.

src/app/public/src/modules/edit-button/edit-button.component.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  private getUrl(): string {
+    const repoData = this.dataService.getByName('repo');
+    const frag = encodeURIComponent(this.routeService.getActiveUrl() + '/');
+    const result = repoData.url + frag + 'index.html';
+
+    return (repoData.vstsBranchSelector === undefined) ? result : result + repoData.vstsBranchSelector;
+  }
+}
```
Blackbaud-SteveBrush: I can't add a comment to the `edit-button.module.spec.ts` file below, but it appears to be empty.

src/app/public/src/modules/wrapper/wrapper.component.ts
```diff
@@ -8,9 +8,11 @@ import { Subscription } from 'rxjs';
 
 import { StacheTitleService } from './title.service';
 import { StachePageAnchorComponent } from '../page-anchor';
-import { StacheJsonDataService } from '../shared';
+import { StacheConfigService, StacheJsonDataService } from '../shared';
 import { StacheNavLink, StacheNavService } from '../nav';
 
+const _get = require('lodash.get');
```
Blackbaud-SteveBrush: Non-blocking feedback:

Adding a new dependency gives me pause, because it 1) adds a feature that we must monitor to prevent introducing a future breaking change, and 2) it increases the risk for transitive dependencies.

In this case, I don't think the cost-benefit of the dependency is merited; I know it offers a lot of convenience for this specific use-case, but I'd rather write out the long-form `if` statement and reduce the risks mentioned.

blackbaud-brandonhare: I'm not sure I agree in this case. This particular method helps eliminate a ton of annoying, repetitive checks in a consistent way. I could see pulling this out into a utils function so that it's centralized, but we constantly do this kind of check. The package itself is small, self contained, and well supported and tested. If you really object that strongly to a dependency, I could see recreating it in a service too instead, but it seems simpler to just pull this in and use it. Thoughts?

Blackbaud-SteveBrush: Sounds good. As long as every other instance where we're doing the long `if` checks are replaced with whatever you decide. Again, non-blocking for this pull request.

src/stache/data/repo.json
```diff
@@ -0,0 +1 @@
+{"url":"https://github.com/blackbaud/stache2/tree/master/src/app/"}
```
Blackbaud-SteveBrush: This still needed?

src/app/public/src/modules/wrapper/wrapper.component.ts
```diff
@@ -94,6 +100,11 @@ export class StacheWrapperComponent implements OnInit, AfterContentInit, OnDestr
     );
   }
 
+  private checkEditButtonUrl(): boolean {
+    const url = _get(this.config, 'skyux.appSettings.stache.editButton.url');
+    return (url !== undefined) ? true : false;
```
Blackbaud-SteveBrush: `(url !== undefined)` already computes to a boolean.

src/app/public/src/modules/wrapper/wrapper.component.ts
```diff
@@ -94,6 +100,11 @@ export class StacheWrapperComponent implements OnInit, AfterContentInit, OnDestr
     );
   }
 
+  private checkEditButtonUrl(): boolean {
+    const url = _get(this.config, 'skyux.appSettings.stache.editButton.url');
```
Blackbaud-SteveBrush: What was @Blackbaud-BrandonJones saying about changing `_get` to something more explanatory? Something like `lodashGet`?

blackbaud-brandonhare: This was the suggestion. Prefixing with the underscore like the lodash package. Since we are just importing one function, I added it to the name instead. `_.get` = `_get`

src/stache/data/repo.json
blackbaud-brandonhare: I still have the updated version of my plugin installed. :P I'll delete this and the plugin and that should stop this from getting added.

src/app/public/src/modules/wrapper/wrapper.component.ts
blackbaud-brandonhare: 👍

src/app/public/src/modules/wrapper/wrapper.component.ts
Blackbaud-BrandonJones: 👍

PR #381, opened 2017-12-18
Added search as explicit dependency.

blackbaud-brandonhare:


package.json
```diff
@@ -36,14 +36,14 @@
   },
   "dependencies": {
     "@blackbaud/skyux-builder-plugin-stache": "~1.3.0",
+    "@blackbaud/skyux-builder-stache-search": "^1.3.0",
```
Blackbaud-SteveBrush: Just to be clear: we want every stache site to have search installed, correct?

blackbaud-brandonhare: Yep!

PR #382, opened 2017-12-20
Added new link component

blackbaud-brandonhare:
I've created a new link component to circumvent the fragment issue. This is the only way to handle it in a consistent manner until Angular figures it out for the framework.


src/app/public/src/modules/link/link.component.html
```diff
@@ -0,0 +1 @@
+<a class="stache-link" (click)="navigate()"><ng-content></ng-content></a>
```
Blackbaud-BrandonJones: With this being an anchor link do we have to do anything with stopping event propagation from the default anchor link click behavior?

blackbaud-brandonhare: We don't do that anyway with nav links. But I can swap it to a div.

blackbaud-brandonhare: I used an anchor tag so I wouldn't have to recreate the styles.

blackbaud-brandonhare: https://github.com/blackbaud/stache2/blob/master/src/app/public/src/modules/nav/nav.component.ts#L35
https://github.com/blackbaud/stache2/blob/master/src/app/public/src/modules/nav/nav.component.html#L6

blackbaud-brandonhare: I think we're fine, especially since we don't have the href on there.

Blackbaud-BrandonJones: I definitely like the anchor link :) It was an actual question lol I don't know if there would be bubbling problems or not.

Blackbaud-BrandonJones: 👍

Blackbaud-BrandonJones:
I like this, care to set up a demo of it working on the app side?  an anchor link to a div with an id or something?

blackbaud-brandonhare:
@Blackbaud-BrandonJones Demo page added to test inter page nav. Also added a demo link on the main demos page so you can test jumps from other pages too.

Blackbaud-SteveBrush:
This works just fine. A few thoughts:

This would be better served as a directive, since it's really only adding functionality, and not a complex template/styles. The directive, `[stacheRouterLink]="route" [fragment]="frag"` could be added to an anchor (or action button, or anything, really), and we wouldn't have to worry about translating every possible attribute for the vanilla `<a>` tag (for example, as it stands, no one would be able to add their own attributes or aria roles: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a). For that to be possible, you'd have to create a 1-to-1 `@Input` for any attributes you wanted to support. With the directive route, you can simply add the `navigate()` method and be done with it.

Blackbaud-SteveBrush:
Related, for future enhancement to this directive: https://github.com/blackbaud/stache2/issues/396

src/app/public/src/modules/link/link.directive.ts
```diff
@@ -0,0 +1,24 @@
...
+  @Input()
+  public stacheRouterLink: string;
+
+  @Input()
+  public fragment: string;
+
+  constructor(private navService: StacheNavService) { }
+
+  @HostListener('click')
+  public navigate(): void {
```
Blackbaud-SteveBrush: Needs an `event.preventDefault();` so that normal anchor linking behavior will be ignored. To test, add `href="#"` to your demo page.

src/app/public/src/modules/link/link.directive.ts
```diff
@@ -0,0 +1,31 @@
...
+  @Input()
+  public fragment: string;
+
+  constructor(
+    private navService: StacheNavService,
+    private el: ElementRef,
+    private renderer: Renderer2
+  ) {
+    this.renderer.setStyle(this.el.nativeElement, 'cursor', 'pointer');
+   }
```
Blackbaud-SteveBrush: I think you have an extra space before the closing curly bracket, here.

src/app/public/src/modules/link/link.directive.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  @HostListener('click', ['$event'])
+  public navigate(event: any): void {
```
Blackbaud-SteveBrush: Go ahead and throw in `MouseEvent` to improve the `event` intellisense.

src/app/public/src/modules/link/link.directive.ts
```diff
@@ -0,0 +1,31 @@
...
+    event.preventDefault();
+    this.navService.navigate({
+      path: this.stacheRouterLink,
+      fragment: this.fragment ? this.fragment : undefined
```
Blackbaud-SteveBrush: The logic on this line could simply be:
`fragment: this.fragment`

src/app/public/src/modules/link/link.directive.ts
blackbaud-brandonhare: Doh!

src/app/public/src/modules/link/link.directive.ts
blackbaud-brandonhare: 👍

src/app/public/src/modules/link/link.directive.ts
blackbaud-brandonhare: I think I was doing something different with this logic initially. Good catch, I'll remove it.

PR #397, opened 2018-01-25
Edit button markup adjustments

Blackbaud-SteveBrush:


src/app/public/src/modules/edit-button/edit-button.component.html
```diff
@@ -1,10 +1,6 @@
-<div class="stache-btn-edit">
-  <button
-    *ngIf="url"
-    type="button"
-    class="sky-btn sky-btn-link"
-    (click)="onEditButtonClick()"
-  >
-    <i class="fa fa-edit"></i> {{ editButtonText }}
-  </button>
+<div class="stache-edit-button" *ngIf="url">
```
Blackbaud-SteveBrush: Class name should mimic the component name.
If the `url` is not set, we should hide the wrapper too.

src/app/public/src/modules/edit-button/edit-button.component.html
```diff
@@ -1,10 +1,6 @@
...
+  <a class="stache-btn-edit" [href]="url">
```
Blackbaud-SteveBrush: Since the action is opening a URL, we should use an anchor, instead of a button.

src/app/public/src/modules/edit-button/edit-button.component.html
```diff
@@ -1,10 +1,6 @@
...
+    <i class="fa fa-edit" aria-hidden="true"></i>
```
Blackbaud-SteveBrush: Hide the icon from screen readers.

src/app/public/src/modules/edit-button/edit-button.component.ts
```diff
@@ -11,14 +10,14 @@ const _get = require('lodash.get');
 })
 export class StacheEditButtonComponent implements OnInit {
   public editButtonText: string;
+  public url: string;
```
Blackbaud-SteveBrush: This property should be public since we're accessing it in the template.

src/app/public/src/modules/edit-button/edit-button.component.ts
```diff
@@ -27,14 +26,10 @@ export class StacheEditButtonComponent implements OnInit {
...
 
-  public onEditButtonClick() {
-    this.navService.navigate({ path: this.url });
-  }
-
   private getUrl(): string {
     const base = _get(this.config, 'skyux.appSettings.stache.editButton.url');
     if (!base) {
-      return;
+      return '';
```
Blackbaud-SteveBrush: My example code failed to return the correct type (needs to be a string).

Issue #377, opened 2017-12-07
component page anchors don't load in TOC


Blackbaud-BrandonJones:
### Expected behavior
If I create a component that has `<stache-page-anchor>`s I would expect them to load on my index file with a table of contents.

### Actual behavior
Table of contents does not pick up these page anchors, they render on the page correctly, but do not display in the table of contents.

### Steps to reproduce
Create a component that has stache page anchors in its template.  Load this component on your index.html page inside a Stache wrapper with showTableOfContents="true".  The anchors will display, the TOC will not pick up the page anchors.

### Solution Possibilities
Create a component that contains <stache-page-anchors> in its template. Then add the component to an index.html page. When rendering that page, the Table of Contents should display the page anchors from the nested component.

I think the solution to this problem is going to require a bigger reworking of how we are handling and passing the data. Likely making use of a page-anchor-service that would track and register all the page anchors. The ToC could reference that service and render accordingly. An official issue will be filed shortly for this work, and hopefully it will make it's way to a sprint before too long.

Blackbaud-BrandonJones:
https://github.com/blackbaud/stache2/pull/441 Resolved

PR #400, opened 2018-01-26
#377 fix page anchor load

Tdortiz:
@Blackbaud-StacyCarlos and I worked on this together and this seems to fix the issue with the page anchors not displaying inside the Table of Contents on page load.


Blackbaud-BrandonJones:
Hey guys thanks for the PR! Though looking at the issue and testing this locally, it does not seem to solve the issue pointed out.

Create a component that contains `<stache-page-anchors>` in its template.  Then add the component to an `index.html` page.  When rendering that page, the Table of Contents should display the page anchors from the nested component.

I think the solution to this problem is going to require a bigger reworking of how we are handling and passing the data.  Likely making use of a page-anchor-service that would track and register all the page anchors.   The ToC could reference that service and render accordingly. An official issue will be filed shortly for this work, and hopefully it will make it's way to a sprint before too long.

If you would like to try your hands at this method, let us know and we can coordinate the effort. :D! We would welcome another pull request!

Blackbaud-BrandonJones:
Being addressed in other PR: https://github.com/blackbaud/stache2/pull/409

PR #409, opened 2018-02-19
Refactor of the page anchor funtionality

Blackbaud-StacyCarlos:
This fixes issue #377 and incorporates an alternate fix for the issue in #400.

Although technically not a breaking change, users would need to add a `name` property to all page anchors to fix a bug. This is unavoidable if the bug is to be completely fixed without resurfacing in the future. This is due to the fact that we don not know when the browser will finishing loading the internal content of the page anchor. Because of this, we don't know when to send the page anchor information to the anchor service. By adding the name property to the inputs of the page anchor, we can know send the information to the anchor service after the component is instantiated.


Blackbaud-BrandonJones:
This is looking good, though there is a slight problem with the Ordering/rendering of anchors that are nested in components.  example

anchor-demo.component.html
```
<stache-page-anchor *ngFor="let t of anchors">
    {{ t.name }}
  </stache-page-anchor>

  <stache-page-anchor>
    {{ changeable }}
  </stache-page-anchor>
```

anchor-demo.component.ts
```
import { Component, OnInit } from '@angular/core';

@Component({
  selector: 'anchor-demo',
  templateUrl: './anchor-demo.component.html'
})
export class StachePageAnchorTestComponent implements OnInit {
  public anchors = [
    {
      name: 'foo'
    },
    {
      name: 'bar'
    },
    {
      name: 'some route'
    }
  ];

  public changeable: string = 'Value One';

  public ngOnInit(): void {
    setTimeout(() => {
      this.changeable = 'different value';
    }, 2000);
  }
}
```

demos/anchor-link/index.html
```
...
  <stache-page-anchor>Test 4</stache-page-anchor>
  <stache-page-anchor>Test 5</stache-page-anchor>
  <anchor-demo></anchor-demo>
  <stache-page-anchor>Test 6</stache-page-anchor>
...
```
Has the following result.  I think this is because when trying to set the name and fragment, It's running before the content is transcluded.  This also has a problem with the order, as the `fragment vs id` isn't set.  
screen shot 2018-03-28 at 8 53 11 am

I think somehow the the wrapper needs to be responsible for querying all the anchors using the `elementRef` and setting the order.  This looks great so far and looks really close :) thanks!

src/app/public/src/modules/wrapper/wrapper.component.spec.ts
```diff
@@ -197,11 +218,11 @@ describe('StacheWrapperComponent', () => {
...
-  it('should have a showTableOfContents input', () => {
-    component.showTableOfContents = true;
-    fixture.detectChanges();
-    expect(component.showTableOfContents).toBe(true);
-  });
+  // it('should have a showTableOfContents input', () => {
+  //   component.showTableOfContents = true;
+  //   fixture.detectChanges();
+  //   expect(component.showTableOfContents).toBe(true);
+  // });
```
Blackbaud-BrandonJones: I forgot to rework and update this test, but it was failing.  We need to either update or remove it.

src/app/public/src/modules/wrapper/wrapper.component.spec.ts
```diff
@@ -287,23 +308,24 @@ describe('StacheWrapperComponent', () => {
...
+  //     return {
+  //       unsubscribe() { }
+  //     };
+  //   });
+  //   const testFixture = TestBed.createComponent(StacheWrapperTestComponent);
+  //   const testComponent = testFixture.componentInstance;
+
+  //   testFixture.detectChanges();
+  //   expect(testComponent.testWrapper.pageAnchorSubscription.length).toEqual(undefined);
+  // });
```
Blackbaud-BrandonJones: See above.

Issue #425, opened 2018-03-20
Remove query string params from breadcrumbs


Blackbaud-BrandonJones:
### Expected behavior
Breadcrumbs should not display the route params
### Actual behavior
Navigating within a Stache page with route params, adds the params to the bread crumbs.
### Steps to reproduce
https://docs.blackbaud.com/engineering-system-docs/standards?foo=bar

breadcrumbs show the params

Blackbaud-LindseyRix:
Reported in Engineering System Slack channel that this issue is affecting page anchors and the ability to select them.

Blackbaud-BrandonJones:
https://github.com/blackbaud/stache2/pull/426

PR #426, opened 2018-03-21
removes query string params from the active url

Blackbaud-BrandonJones:
https://github.com/blackbaud/stache2/issues/425

This fix will remove the Query String params from showing up in the bread crumbs. 

screen shot 2018-03-21 at 8 06 50 am


src/app/public/src/modules/nav/nav.service.ts
```diff
@@ -10,18 +10,19 @@ export class StacheNavService {
...
+    const currentPath = this.router.url.split('?')[0].split('#')[0].substring(1);
+
     if (this.isExternal(route)) {
       this.windowRef.nativeWindow.location.href = route.path;
       return;
     }
 
     if (route.fragment) {
-      const element = this.windowRef.nativeWindow.document.getElementById(route.fragment);
+      const element = this.windowRef.nativeWindow.document.querySelector(`#${route.fragment}`);
```
Blackbaud-SteveBrush: Curious why you went with `querySelector`? It's a bit slower than `getElementById`.

Blackbaud-BrandonJones: I went with `querySelector` because most examples I have seen were posted using it.  It seemed like a more standard practice.

PR #431, opened 2018-03-28
added omnibar service and method to wrapper

Blackbaud-BrandonJones:
So the omnibar service was only ever adding the needed class to the body if it was instantiated. This was only happening if there was an instance of `affix-top` on the page.  So until that directive was used, the page anchors would not scroll properly with the omnibar present.  This fix ensures that the service checks for the omnibar and adds the appropriate class on any stache page.


src/app/public/src/modules/shared/omnibar-adapter.service.ts
```diff
@@ -7,9 +7,13 @@ export class StacheOmnibarAdapterService {
...
   public static readonly EXPECTED_OMNIBAR_HEIGHT: number = 50;
   private element: any = this.windowRef.nativeWindow.document.querySelector('.sky-omnibar-iframe');
+  private configured: boolean = false;
 
-  constructor(private windowRef: StacheWindowRef) {
-    if (this.omnibarEnabled()) {
+  constructor(private windowRef: StacheWindowRef) { }
+
+  public checkForOmnibar(): void {
+    if (this.omnibarEnabled() && !this.configured) {
```
Blackbaud-SteveBrush: LGTM. I would have personally rewired this class to use `Renderer2` to add the class name since it would only ever add the class once (and you'd then avoid the `this.configured` check), but I know you're spent for time.

Blackbaud-SteveBrush: Just by way of demonstration:

```
import { Injectable, Renderer2, RendererFactory2 } from '@angular/core';

import { StacheWindowRef } from './window-ref';

@Injectable()
  private renderer: Renderer2;

  constructor(
    private rendererFactory: RendererFactory2,
    private windowRef: StacheWindowRef
  ) {
    this.renderer = this.rendererFactory.createRenderer(undefined, undefined);
  }

  public checkForOmnibar(): void {
    if (this.omnibarEnabled()) {
      this.applyClassToBody();
    }
  }

  public getHeight(): number {
    return (this.omnibarEnabled()) ? 50 : 0;
  }

  public omnibarEnabled(): boolean {
    return !!this.windowRef.nativeWindow.document.querySelector('.sky-omnibar-iframe');
  }

  private applyClassToBody(): void {
    this.renderer.addClass(
      this.windowRef.nativeWindow.document.body,
      'stache-omnibar-enabled'
    );
  }
}
```

But never mind; merge away!

Blackbaud-BrandonJones: I will implement this :) I totally should have thought of that.

Issue #446, opened 2018-04-18
Double sidebar scrollbars


Blackbaud-BrandonJones:
### Expected behavior
When the side-bar affixes to the top it should only have one scrollbar when it's longer than the page. 
### Actual behavior
Two scrollbars can be seen.
### Steps to reproduce
serve Stache2 and look at the sidebar on the home page when scrolling down.
![screen shot 2018-04-18 at 9 28 31 am](https://user-images.githubusercontent.com/20796436/38934931-06b41a22-42eb-11e8-873f-13235ca2bb83.png)

Blackbaud-LindseyRix:
Fixed.

Issue #424, opened 2018-03-19
Switching between pages causes shifting of sidebar nav + breadcrumbs


Blackbaud-LindseyRix:
On the learn > getting started > tutorials....note that if you toggle between tutorial 1 and tutorial 2, the page margins jump around so that the page contents section shifts to the left a bit.  

https://docs.blackbaud.com/uiextensibility-docs/learn/tutorials/tutorial-two

Lindsey has an animated gif of it happening. GH is not letting me upload files from work so will attach to backlog item.

Reported by: Ben Lambert
- Chrome latest version - looks to be 64.0.3282.186

Blackbaud-LindseyRix:
Per @Blackbaud-BrandonJones, it’s only on windows machines in chrome, it is the scroll bar showing up.

Blackbaud-LindseyRix:
When observing this happen on Ben's computer, it appears that this happens when the page goes from being within the view to requiring a user to scroll down the page. When that happens, the page scrollbar appears on the right (previously hidden), and causes the content to shift on the page. 
- Happens on Windows, Chrome
- Doesn't happen on Mac, Chrome

Blackbaud-BrandonJones:
This was addressed in release 2.8.0 https://github.com/blackbaud/stache2/pull/486
Ben Lambert has been contacted and notified.

PR #447, opened 2018-04-18
Bugfix jumping sidebar

Blackbaud-BrandonJones:
Resolves both:
issue https://github.com/blackbaud/stache2/issues/424
and
issue https://github.com/blackbaud/stache2/issues/446


src/app/public/src/modules/wrapper/wrapper.component.scss
```diff
@@ -1,5 +1,15 @@
 @import '../../styles/_variables';
 
+/*
+* We add this overflow-y so that when switching between screens with and without a vertical
+* scroll bar the page doesn't "jump" 16px back and forth.
+* https://stackoverflow.com/a/7607206
```
Blackbaud-SteveBrush: "Worth mentioning that this permanently shows a vertical scrollbar. May not always be acceptable." :(

Blackbaud-SteveBrush:
I'm struggling to see the importance of fixing #424? It seems that that is just how browser pages work and I wouldn't want to mess with overflow on the html tag...

Blackbaud-LindseyRix:
The developer who reported it is Ben Lambert. I take it seriously the requests he makes and would appreciate it if we fix this for him.

Blackbaud-BrandonJones:
This is being resolved with the reformatting of the sidebar layout.

Issue #412, opened 2018-02-22
Browser page title not set if pageTitle isn't set on Stache tag


Blackbaud-LindseyRix:
### Expected behavior
Can we provide back-up methods for setting browser page title from other properties if pageTitle isn't set? navTitle?

### Actual behavior:
Looks like we are pulling Browser page title from the pageTitle property on Stache tag. Some pages don't have a pageTitle (like tutorials).  When that happens, we only populate the name of the site, circumventing our efforts to make sure every page has a browser title for identification in Analytics. 

https://developer.blackbaud.com/stache/learn/get-started/add-stache

PR #448, opened 2018-04-18
Update Browser Page Title functionality

Blackbaud-TimPepper:
Include navTitle and tutorialHeader as options for browser page title when pageTitle stache attribute is absent

### DIFF
--- a/src/app/public/src/modules/wrapper/title.service.ts
+++ b/src/app/public/src/modules/wrapper/title.service.ts
@@ -14,7 +14,7 @@ export class StacheTitleService {
 
     if (parts && parts.length > 0) {
       parts.push(windowTitle);
-      let validParts = parts.filter(part => part !== undefined);
+      let validParts = parts.filter(part => part);
       windowTitle = validParts.join(' - ');
     }
 
--- a/src/app/public/src/modules/wrapper/wrapper.component.spec.ts
+++ b/src/app/public/src/modules/wrapper/wrapper.component.spec.ts
@@ -37,6 +37,7 @@ describe('StacheWrapperComponent', () => {
   let mockWindowService: any;
   let mockAnchorService: any;
   let mockOmnibarService: any;
+  let mockTextContent: string = '';
 
   class MockActivatedRoute {
     public fragment: Observable<string> = Observable.of('test-route');
@@ -112,6 +113,7 @@ describe('StacheWrapperComponent', () => {
         }),
         querySelector: jasmine.createSpy('querySelector').and.callFake(function(selector: string) {
           return {
+            textContent: mockTextContent,
             classList: {
               add(cssClass: string) { }
             },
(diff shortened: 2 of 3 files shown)


Blackbaud-TimPepper:
https://github.com/blackbaud/stache2/issues/412

src/app/public/src/modules/wrapper/title.service.ts
```diff
@@ -14,7 +14,7 @@ export class StacheTitleService {
 
     if (parts && parts.length > 0) {
       parts.push(windowTitle);
-      let validParts = parts.filter(part => part !== undefined);
+      let validParts = parts.filter(part => part);
```
Blackbaud-BrandonJones: I think we would want to make this a little more verbose for readability. It's not immediately apparent why we would be trying to call a filter and just return the part being passed in.  
```
parts.filter(part => (part !== undefined && part !== ''))
```
thoughts? @blackbaud-brandonhare @Blackbaud-SteveBrush ?

src/app/public/src/modules/wrapper/wrapper.component.ts
```diff
@@ -66,16 +65,18 @@ export class StacheWrapperComponent implements OnInit, OnDestroy, AfterViewInit
...
 
   public ngOnInit(): void {
     this.omnibarService.checkForOmnibar();
-    this.titleService.setTitle(this.windowTitle || this.pageTitle);
     this.jsonData = this.dataService.getAll();
     this.registerPageAnchors();
   }
 
   public ngAfterViewInit() {
+    this.checkForTutorialHeader();
```
Blackbaud-BrandonJones: I like this idea, but perhaps we can clean up this code some by calling another method.  Maybe a `getTitle()` method that makes the checks and returns the preferred value?  This would clean up the variables being passed into that `setTitle` method.

src/app/public/src/modules/wrapper/title.service.ts
Blackbaud-SteveBrush: Yeah, I agree; it's a bit difficult to gauge the purpose of the `filter` at first glance.

src/app/public/src/modules/wrapper/wrapper.component.ts
```diff
@@ -54,6 +52,7 @@ export class StacheWrapperComponent implements OnInit, OnDestroy, AfterViewInit
   @Input()
   public showBackToTop: boolean = true;
 
+  public tutorialHeader: string;
```
Blackbaud-SteveBrush: Is there a reason this variable needs to be public?

src/app/public/src/modules/wrapper/wrapper.component.ts
Blackbaud-SteveBrush: I like that, @Blackbaud-BrandonJones

src/app/public/src/modules/wrapper/title.service.ts
Blackbaud-TimPepper: This was the only change I wasn't super sure about. I think that we can be comfortable saying that anything falsey shouldn't make it through the filter. Adding more conditions just makes the filter more complex, and could potentially let through falsey values we aren't expecting in the future. I would be more comfortable adding a comment explaining the filter, in this instance. Let me know your thoughts!

src/app/public/src/modules/wrapper/wrapper.component.spec.ts
```diff
@@ -287,12 +289,50 @@ describe('StacheWrapperComponent', () => {
...
+    mockTextContent = 'Tutorial Header Title';
+    component.ngOnInit();
+    fixture.detectChanges();
+    expect(mockTitleService.setTitle).toHaveBeenCalledWith('Tutorial Header Title');
+    mockTextContent = '';
+  });
+
+  it('should not set the window title when the Tutorial Header title is absent', () => {
+    fixture.detectChanges();
+    expect(mockTitleService.setTitle).not.toHaveBeenCalledWith('Tutorial Header Title');
```
Blackbaud-BrandonJones: This test doesn't really tell us much except it wasn't called with `Tutorial Header Title`.  What was it called with? Anything?  What gets passed in if there is no pageTitle, navTitle, or tutorial header? etc.

src/app/public/src/modules/wrapper/wrapper.component.ts
```diff
@@ -66,16 +64,18 @@ export class StacheWrapperComponent implements OnInit, OnDestroy, AfterViewInit
...
   public ngOnInit(): void {
     this.omnibarService.checkForOmnibar();
-    this.titleService.setTitle(this.windowTitle || this.pageTitle);
     this.jsonData = this.dataService.getAll();
     this.registerPageAnchors();
   }
 
   public ngAfterViewInit() {
+    const preferredWindowHeader = this.setWindowHeader();
+    this.titleService.setTitle(preferredWindowHeader);
```
Blackbaud-BrandonJones: Much cleaner! 👍

Issue #415, opened 2018-03-02
Enable routerLink option to right-click open in new window


Blackbaud-LindseyRix:
### Expected behavior

RouterLinks in Angular aren't treated like hrefs, so we take away the option for a user to right-click open a link in a a new tab/window.

### Actual behavior

If we use stacheRouterLink, or even routerLink, we don't give the user the option to open a link in a new tab/window if they don't want to leave the page they are on. For external links, we can change stacheRouterLink to hrefs, but that still doesn't solve the problem for routerLinks within a site. 

This has been requested by Kim Coughlin, Derek Nichols, and RoxAnna Fuquay. They noted they would like it for action buttons, in addition to links.

Items Requested to have option for open in new window/browser tab:
- `stacheRouterLink`
- Action buttons

We should also add a documentation best practice note to the Tutorial component, that we recommend users enable links to be opened in a new window from a tutorial. When someone is following a tutorial, it is recommended that they give users the option to open in a new window/tab so they don't lose their step in the process.

Blackbaud-TimPepper:
Users should now be able to use the StacheRouterLink with as much functionality as RouterLink, and then some. StacheRouterLink should now be used primarily, instead of RouterLink. 

Functionality includes: ctrl/shift/cmd(meta) + click navigation, and browser right-click context menu options, which were missing previously with StacheRouterLink.

PR #458, opened 2018-05-03
Update stache router link usability

Blackbaud-TimPepper:
https://github.com/blackbaud/stache2/issues/415


Blackbaud-TimPepper:
This PR aims to replace the angular RouterLink directive completely with the StacheRouterLink directive, with all the same functionality, and more. This would allow us to be consistent with which directive we are using. The changes themselves allow the user to right-click to view the browser context window options to open a new tab or window. The user can also ctrl/shift/meta(command for mac)+click to open a new tab or window with the link.

src/app/public/src/modules/nav/nav.component.ts
```diff
@@ -74,7 +76,7 @@ export class StacheNavComponent implements OnInit, StacheNav {
       path = `/${path}`;
     }
 
-    const isActiveParent = (navDepth > 1  && `${activeUrl}/`.indexOf(`${path}/`) === 0);
```
Blackbaud-BrandonJones: good catch!

src/app/public/src/modules/nav/nav.component.ts
```diff
@@ -32,8 +32,10 @@ export class StacheNavComponent implements OnInit, StacheNav {
     return Array.isArray(route.children);
   }
 
-  public navigate(route: any): void {
-    this.navService.navigate(route);
+  public navigate(event: KeyboardEvent, route: any): void {
```
Blackbaud-BrandonJones: is this still necessary if we are using the `StacheLinkDirective` ?

Blackbaud-TimPepper: I'll look into this, but you're probably right.

src/app/public/src/modules/nav/nav.component.html
```diff
@@ -4,10 +4,10 @@
       *ngFor="let route of routes"
       [ngClass]="{'active': route.isActive, 'current': route.isCurrent}">
       <a class="stache-nav-anchor"
-        routerLink="{{route.path}}"
+        stacheRouterLink="{{route.path}}"
         fragment="{{route.fragment}}"
         queryParamsHandling="merge"
-        (click)="navigate(route)">{{route.name}}</a>
+        (click)="navigate($event, route)">{{route.name}}</a>
```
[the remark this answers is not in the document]
Blackbaud-TimPepper: This and the below instance are both necessary because we need to stop the component from navigating to the anchor on the initial page if we open in a new tab/window.

Blackbaud-TimPepper:
Update: This all looks good in IE11

src/app/public/src/modules/link/link.directive.ts
```diff
@@ -1,31 +1,73 @@
...
+
   @HostListener('click', ['$event'])
   public navigate(event: MouseEvent): void {
     event.preventDefault();
-    this.navService.navigate({
-      path: this.stacheRouterLink,
-      fragment: this.fragment
-    });
+    if (event.ctrlKey || event.metaKey || event.shiftKey) {
+      this.windowRef.nativeWindow.open(this.href);
```
Blackbaud-SteveBrush: May want to avoid `open`, see: https://developer.mozilla.org/en-US/docs/Web/API/Window/open#Usability_issues

Blackbaud-SteveBrush: Another thought, the nav service is already checking to see if the `path` is "global"; maybe you could tie into that somehow? https://github.com/blackbaud/stache2/blob/master/src/app/public/src/modules/nav/nav.service.ts#L17

(I created a `stacheRouterLink` on the demo page that pointed to an external URL and it seemed to work.)

Blackbaud-TimPepper: After doing a little bit of research it seems like window.open is really the only way to open a new tab/window in the browser. A lot of the warnings given in the mozilla link are in reference to issues that have since been resolved by native browser functionality, for instance: chrome and IE will automatically open a new tab instead of a browser window when ctrl+clicking. I believe that can be changed in the settings if the user would rather open a window. 

As far as I can tell there are no other alternatives that would provide us with the same level of functionality as the routerLink module. Let me know if you have more insight!

Blackbaud-BrandonJones:
```
  @HostListener('click', ['$event'])
  public navigate(event: MouseEvent): boolean {
    if (event.ctrlKey || event.metaKey || event.shiftKey) {
      return true;
    }

    this.navService.navigate({
      path: this.stacheRouterLink,
      fragment: this.fragment
    });
    return true;
  }
```
If we don't `preventDefault()` and we return true, it seems to allow the browser default actions for the anchor link.  This might let us avoid using `window.open`.  Is there any issue with doing this?  I think that's how Angular is handling it.

https://github.com/angular/angular/blob/5.2.10/packages/router/src/directives/router_link.ts#L217
https://github.com/angular/angular/blob/5.2.10/packages/router/src/directives/router_link.ts#L144

src/app/public/src/modules/nav/nav.component.html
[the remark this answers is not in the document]
Blackbaud-BrandonJones: Not sure I follow.  Removing the click event entirely from this component seems to work fine when opening in a new tab, or window, or clicking directly on the page.  Am I missing something?

src/app/public/src/modules/nav/nav.component.ts
Blackbaud-BrandonJones: Removing the `click` event from the `nav.component.html` this change may not be necessary :)

Blackbaud-TimPepper: We need the event to stop the page from navigating internally when we ctrl+click. Without the event it opens a new tab AND navigates on the current page.

src/app/public/src/modules/link/link.directive.ts
Blackbaud-TimPepper: @Blackbaud-BrandonJones 's approach seems to work. The preventDefault still needs to go outside of the if statement though. It's what is stopping the page from reloading every time you navigate to an in-page link.

src/app/public/src/modules/nav/nav.component.html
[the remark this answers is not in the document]
Blackbaud-TimPepper: We definitely still need the event in the component, but it looks like removing it here still feeds the event to the function, if that's what you mean. I felt that leaving the event made the workflow more readable.

src/app/public/src/modules/link/link.directive.ts
Blackbaud-BrandonJones: I can't seem to reproduce this?  With the return false at the end it won't reload.  Also which link are you clicking on to see this action happen?  TOC or links using the `nav.component` would need to have the `click` event removed from the component, otherwise it causes the reload?

src/app/public/src/modules/link/link.directive.ts
```diff
@@ -1,31 +1,74 @@
...
+
   @HostListener('click', ['$event'])
   public navigate(event: MouseEvent): void {
-    event.preventDefault();
-    this.navService.navigate({
-      path: this.stacheRouterLink,
-      fragment: this.fragment
-    });
+    if (event.ctrlKey || event.metaKey || event.shiftKey) {
+      return;
```
Blackbaud-BrandonJones: I think we want to be explicit and return `true` here so the events continue to propagate. returning undefined can be interpolated as a false value.  Thoughts @Blackbaud-SteveBrush @blackbaud-brandonhare ?

Blackbaud-BrandonJones:
I think this looks good to me @Blackbaud-SteveBrush ? @blackbaud-brandonhare ?

src/app/public/src/modules/link/link.directive.ts
```diff
@@ -1,31 +1,75 @@
-import { Directive, ElementRef, HostListener, Input, Renderer2 } from '@angular/core';
+import { Directive, ElementRef, HostListener, HostBinding, Input, Renderer2, OnChanges, AfterViewInit } from '@angular/core';
```
blackbaud-brandonhare: I'd wrap these.

blackbaud-brandonhare:
Other than wrapping the imports in that one file, this looks good to me.

PR #460, opened 2018-05-09
Update changelog and version

Blackbaud-TimPepper:


CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+# 2.7.0 (2018-5-9)
+
+- StacheRouterLink bugfixes [#452](https://github.com/blackbaud/stache2/pull/452)
+  - Fixed issues that stopped the page from navigating if fragment existed on current and other page
```
Blackbaud-BrandonJones: This is more specifically it would not navigate to the fragment on another page, if the current page also contained a fragment with the same name.

PR #462, opened 2018-05-11
using fa icon instead of svg

Blackbaud-BrandonJones:
Builder does not allow libraries to pull in their own assets, so until that gets resolved we are using a font awesome icon.


blackbaud-brandonhare:
Didn't we do it this way initially because it wouldn't work on IE11? Does it still work as originally intended?

Blackbaud-BrandonJones:
Before it was an inline SVG string loaded into the element. This is just using an FA icon

blackbaud-brandonhare:
Right, but didn't we have it that way before and it didn't work for some reason? Maybe @Blackbaud-SteveBrush can speak more to that. I could be misremembering.

Blackbaud-BrandonJones:
tested and it works in IE

PR #466, opened 2018-05-22
Feature footer component

Blackbaud-BrandonJones:
Proposed/sample style changes to sidebar layout, and footer component.


src/app/public/src/modules/footer/footer.component.ts
```diff
@@ -0,0 +1,30 @@
...
+  styleUrls: ['./footer.component.scss']
+})
+export class StacheFooterComponent implements OnInit {
+  public copyright: string;
+  public siteName: string;
+  public resourceLinks: StacheNavLink[];
+
+  constructor(private configService: StacheConfigService) { }
+
+  ngOnInit() {
```
Blackbaud-TimPepper: Need an access modifier here

Blackbaud-BrandonJones: Yeah for sure, the code in this PR is not ready for review, it's mostly non functional at the moment anyway.  This was just to make the branch/style changes easier to see to get feedback. :) Thanks and good catch though!

blackbaud-brandonhare:
I like where this is going. Makes better use of the available space. I would say I think I would like it better if the footer went all the way across the bottom of the page, but that's just my opinion.

Blackbaud-BrandonJones:
This mimics the sidebar pattern angular uses
https://angular.io/docs
The footer going the full page width causes a lot of issues because of our affixed sidebar.

blackbaud-brandonhare:
Yeah, it's more of an opinion from me, so take it with a grain of salt. But they have the sidebar on every page. How will that translate to all of the different patterns in Stache?

Blackbaud-BrandonJones:
to clarify: The footer content and code itself is not set or finished by any means yet.  It's simply holding space.  This PR at the moment is just to display the proposed sidebar changes.

blackbaud-brandonhare:
I like where the sidebar is going then. We can shelve my footer comments until later. :)

Blackbaud-BrandonJones:
```
RE: Yeah, it's more of an opinion from me, so take it with a grain of salt. But they have the sidebar on every page. How will that translate to all of the different patterns in Stache?
```
It spans the entire page content area as expected.  You can see a demo of this on the tutorial 2 page under Demos.

Blackbaud-TimPepper:
I think the issue with Affix is purely an Affix issue, to that point. If there's an issue with how it behaves we shouldn't let it get in the way of us making a quality footer that we can be proud of. We should probably put an issue in to look in to why it's behaving undesirably and fix it so it doesn't hang us up on layout updates in the future.

blackbaud-brandonhare:
Yeah, the footer content jumps depending on what type of page you are on. I think it would be less impactful if you either had a consistent footer throughout OR if you centered the content. But I would prefer the consistent footer personally.

Blackbaud-TimPepper:
`Re: It spans the entire page content area as expected. You can see a demo of this on the tutorial 2 page under Demos.`

That looks good. If we flattened and centered the content it would probably be really nice

Blackbaud-TimPepper:
What if we did something like this: 

image

image

Not saying it looks better or worse, but just a suggestion.

Also, I like that the content in your example is justified to the center of the footer area, but the text itself needs to be formatted in a different way I think.

Blackbaud-LindseyRix:
Is there no way on the sidebar layout to have the footer be full page width so it's always consistent at the bottom of the page? Also, have we found any other sites that provide a best practice model for these?

Blackbaud-BrandonJones:
The footer is full page, the sidebar just sits over the top of it. This is being modeled from Angular's docs, and next phase will to make that sidebar collapsible.
https://angular.io/docs

Re formatting text:  The links need to be stacked because they will be an input and we don't know how  many footer links users will want to put there.

Blackbaud-SteveBrush:
@Blackbaud-BrandonJones @Blackbaud-TimPepper Would it be possible to get a pull request that includes only the footer changes? It's fairly difficult to piece out what's a change and what's a "what-if-the-sidebar-gets-merged" change...

src/app/public/src/modules/footer/footer.component.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  public ngOnInit(): void {
+    this.currentDate = new Date();
+    this.setFooterData();
+  }
+
+  private setFooterData(): void {
+    this.footerLinks = _get(this.configService, 'skyux.appSettings.stache.footer.nav');
+    this.copyright = _get(this.configService, 'skyux.appSettings.stache.footer.copyright', `Blackbaud, Inc. All rights reserved.`);
+    this.siteName = _get(this.configService, 'skyux.name');
```
Blackbaud-SteveBrush: This should probably point to `app.title` instead?

Blackbaud-BrandonJones: Is that a more reliable field?

skyuxconfig.json
```diff
@@ -37,6 +37,18 @@
       },
       "searchConfig": {
       },
+      "footer": {
+        "nav": [
+          {
+            "title": "Privacy Policy",
```
Blackbaud-SteveBrush: Should we use `name` instead of `title` to keep it consistent with the existing nav service?
https://github.com/blackbaud/stache2/blob/master/src/app/public/src/modules/nav/nav-link.ts#L2

src/app/public/src/modules/container/container.module.ts
```diff
@@ -0,0 +1,18 @@
...
+  ],
+  declarations: [
+    StacheContainerComponent
+  ],
+  exports: [
+    StacheContainerComponent
+  ],
+  providers: []
+})
+export class StacheContainerModule { }
```
Blackbaud-SteveBrush: Moving `StacheContainerComponent` to a new directory is a breaking change. What is the goal for moving it to its own module?

src/app/public/src/modules/footer/footer.component.html
```diff
@@ -0,0 +1,12 @@
+<div class="stache-footer-wrapper">
+  <div class="stache-footer-content">
+    <h3 *ngIf="siteName" class="stache-footer-site-name">
+      <strong>{{ siteName }}</strong>
```
Blackbaud-SteveBrush: I would remove this extraneous `strong` tag and just add styles to the `stache-footer-site-name` class.

src/app/public/src/modules/footer/footer.component.html
```diff
@@ -0,0 +1,12 @@
...
+    </h3>
+    <stache-nav navType="footer" [routes]="footerLinks">
+    </stache-nav>
+    <div *ngIf="copyright" class="copyright-label">
```
Blackbaud-SteveBrush: Mind prefixing the class with `stache-footer-`?

src/app/public/src/modules/container/container.module.ts
Blackbaud-BrandonJones: Originally this was moved because the footer was also going to make use of the container, that has changed so it can be moved back.

skyuxconfig.json
Blackbaud-BrandonJones: I think we talked and wanted this to match the nav pattern from the omnibar?  I am fine either way, what's easiest for our users?

Blackbaud-BrandonJones:
Reworked with : https://github.com/blackbaud/stache2/pull/489

PR #469, opened 2018-05-29
[HOLD] Add copy to clipboard directive

Blackbaud-TimPepper:
Adding a directive so people can copy content to their clipboard. Mostly to be used with code blocks. 

Please provide any and all feedback! It's still WIP, so there's a lot that can be improved. Biggest thing to note so far: the copy-to-clipboard button is being injected into the DOM after the element loads, and then positioned in the top right of the parent element. To accomplish this I'm currently setting the parent element position to relative, so I can position the button at `position: absolute; top: 0; right: 0`. I'd like to not have to fiddle with the parent element, as that could have potential adverse effects. If anyone is a CSS guru this is your time to shine.


src/app/public/src/modules/clipboard/clipboard.service.ts
```diff
@@ -0,0 +1,104 @@
...
+import { Inject, Injectable, Renderer } from '@angular/core';
+import { DOCUMENT } from '@angular/platform-browser';
+import { WINDOW } from 'ngx-window-token';
+
+@Injectable()
+export class ClipboardService {
+    private tempTextArea: HTMLTextAreaElement | undefined;
+    constructor(
+        @Inject(DOCUMENT) private document: any,
+        @Inject(WINDOW) private window: any,
```
Blackbaud-BrandonJones: I think we can use our `windowRef` here instead of installing a library that does the same thing?
https://github.com/blackbaud/stache2/blob/master/src/app/public/src/modules/affix/affix-top.directive.ts#L9

Blackbaud-BrandonJones:
`If anyone is a CSS guru this is your time to shine.` - @Blackbaud-SteveBrush Your time to shine, is now :D!

PR #476, opened 2018-06-05
Add copy to clipboard directive and CodeBlock implementation

Blackbaud-TimPepper:
Looking for review around the follow changes:

The work for the clipboard component has been removed with the intent of coming back to it in the future. Now the entirety of the work is focused around the copy-to-clipboard functionality in the clipboard directive/service and a simple implementation given in the Stache Code-Block component. For the time being this is the most urgent use case, and building a Stache component to handle a basic use case will require a significant amount of design work that we hope to take on in the future. 

An example of the working code can be found in the Code Block are of the Demos.


src/app/public/src/modules/clipboard/clipboard.component.html
```diff
@@ -0,0 +1,8 @@
+<div class="clipboard-wrapper" stacheClipboard clipboardTrigger="{{clipboardTrigger}}">
+  <ng-content></ng-content>
+  <button
+    *ngIf="chooseTrigger()"
+    class="sky-btn sky-btn-default clipboard-btn"
```
Blackbaud-SteveBrush: Button needs a `type="button"` so it doesn't submit forms.

src/app/public/src/modules/clipboard/clipboard.component.html
```diff
@@ -0,0 +1,8 @@
+<div class="clipboard-wrapper" stacheClipboard clipboardTrigger="{{clipboardTrigger}}">
```
Blackbaud-SteveBrush: Prefix classes with `stache-`.

src/app/public/src/modules/clipboard/clipboard.component.html
```diff
@@ -0,0 +1,8 @@
...
+  <ng-content></ng-content>
+  <button
+    *ngIf="chooseTrigger()"
+    class="sky-btn sky-btn-default clipboard-btn"
+    id="clipboardButton">
```
Blackbaud-SteveBrush: Avoid using `id` within a reusable Angular component, since there could be multiple elements on the page with the same ID.

src/app/public/src/modules/clipboard/clipboard.component.ts
```diff
@@ -0,0 +1,52 @@
...
+      this.clipboardButton.textContent = this.buttonText;
+    } else {
+      this.clipboardButton = <HTMLButtonElement>document.querySelector(`#${this.clipboardTrigger}`);
+    }
+
+    if (!this.customClickHandler) {
+      const buttonText = this.clipboardButton.textContent;
+      this.renderer.listen(this.clipboardButton, 'click', () => {
+        this.clipboardButton.textContent = this.copyText;
+        setTimeout(() => { this.clipboardButton.textContent = buttonText; }, 1000);
```
Blackbaud-SteveBrush: Reference `setTimeout` from the `StacheWindowService`.

src/app/public/src/modules/clipboard/clipboard.component.ts
```diff
@@ -0,0 +1,52 @@
+import { Component, Input, ElementRef, AfterViewInit, Renderer2 } from '@angular/core';
+
+@Component({
+  selector: 'stache-clipboard',
+  templateUrl: './clipboard.component.html',
+  styleUrls: ['./clipboard.component.scss']
+})
+export class StacheClipboardComponent implements AfterViewInit {
+  private defaultTigger = 'clipboardButton';
+  private clipboardButton: HTMLButtonElement;
```
Blackbaud-SteveBrush: You can probably get away with using `ViewChild` instead of referencing the DOM directly.
https://angular.io/api/core/ViewChild

src/app/public/src/modules/clipboard/clipboard.component.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  @Input()
+  public customClickHandler: boolean = false;
+
+  @Input()
+  public buttonText: string = 'Copy to Clipboard';
```
Blackbaud-SteveBrush: These English strings should be referenced from `SkyAppResourcesService`.
https://developer.blackbaud.com/skyux/learn/get-started/advanced/localization

src/app/public/src/modules/clipboard/clipboard.directive.spec.ts
```diff
@@ -0,0 +1,127 @@
+// import { By } from '@angular/platform-browser';
```
Blackbaud-SteveBrush: Why commented-out?

src/app/public/src/modules/clipboard/clipboard.module.ts
```diff
@@ -0,0 +1,38 @@
...
+import { DOCUMENT } from '@angular/platform-browser';
+
+import { ClipboardService } from './clipboard.service';
+import { StacheClipboardComponent } from './clipboard.component';
+import { FormsModule } from '@angular/forms';
+import { StacheClipboardDirective } from './clipboard.directive';
+import { StacheWindowRef } from '../shared';
+
+// this pattern is mentioned in https://github.com/angular/angular/issues/13854 in #43
+export function clipboardServiceFactory(doc: Document,
```
Blackbaud-SteveBrush: Needs a `stache` prefix.

src/app/public/src/modules/clipboard/clipboard.service.ts
```diff
@@ -0,0 +1,92 @@
+// adapted from ngx-clipboard
+
+import { Inject, Injectable, Renderer2 } from '@angular/core';
+import { DOCUMENT } from '@angular/common';
+import { StacheWindowRef } from '../../..';
+
+@Injectable()
+export class ClipboardService {
```
Blackbaud-SteveBrush: Needs a `Stache` prefix.

src/app/public/src/modules/clipboard/clipboard.directive.ts
```diff
@@ -0,0 +1,80 @@
...
+export class StacheClipboardDirective implements AfterViewInit, OnDestroy {
+  // --------------------------------------------------------------
+  // NOTE: Only supported for Code-Block component as of July 2018
+  // --------------------------------------------------------------
+
+  private defaultTigger = 'clipboardButton';
+
+  @Input() public clipboardTrigger: string = this.defaultTigger;
+  @Input() public copySuccessText: string;
+  @Output() public cbOnSuccess: EventEmitter<any> = new EventEmitter<any>();
```
Blackbaud-SteveBrush: The word "on" is discouraged by Angular style guide: https://angular.io/guide/styleguide#dont-prefix-output-properties.

We generally name events that happen multiple times in the present tense, and events that occur only once in the past tense.

Also, are these outputs even needed at this point? I tend to "under" release features to avoid complexity, and only add features when they're directly asked for. Since this directive is internal-only, we can avoid overcomplicating the API until it's released to the public.

src/app/public/src/stache.module.ts
```diff
@@ -42,6 +43,7 @@ export * from './modules/shared';
     StacheBackToTopModule,
     StacheBlockquoteModule,
     StacheBreadcrumbsModule,
+    StacheClipboardModule,
```
Blackbaud-SteveBrush: I would not export the clipboard module to the public, since it's only being used internally by the code block.

src/app/public/src/modules/code-block/code-block.component.ts
```diff
@@ -76,6 +77,13 @@ export class StacheCodeBlockComponent implements AfterViewInit {
     this.cdRef.detectChanges();
   }
 
+  public handleClipboardButton() {
+    this.clipboardButtonClicked = true;
+    setTimeout(() => {
```
Blackbaud-SteveBrush: Is there a more "graceful" way to handle this? What happens if I spammed the copy button?

It would also be nice to show a "Copied!" message near the button, much like this demo:
https://clipboardjs.com/
screen shot 2018-07-05 at 12 13 31 pm

Maybe that could be accomplished with the SKY UX popover?
https://developer.blackbaud.com/skyux/components/popover

src/app/public/src/modules/code-block/code-block.component.html
```diff
@@ -1,6 +1,16 @@
...
+  <div #codeFromContent class="stache-code-input">
+    <ng-content></ng-content>
+  </div>
+  <div stacheClipboard clipboardTrigger="copyCodeButton">
+    <pre class="stache-code-output"><code [ngClass]="getClassName()" [innerHTML]="output"></code></pre>
+    <button
+      aria-label="copy to clipboard"
+      class="sky-btn sky-btn-default stache-clipboard-btn"
+      id="copyCodeButton"
+      (click)="handleClipboardButton()">
```
Blackbaud-SteveBrush: Button needs `type="button"`.

src/app/public/src/modules/code-block/code-block.component.html
```diff
@@ -1,6 +1,16 @@
...
\ No newline at end of file
+  <div #codeFromContent class="stache-code-input">
+    <ng-content></ng-content>
+  </div>
+  <div stacheClipboard clipboardTrigger="copyCodeButton">
+    <pre class="stache-code-output"><code [ngClass]="getClassName()" [innerHTML]="output"></code></pre>
+    <button
+      aria-label="copy to clipboard"
+      class="sky-btn sky-btn-default stache-clipboard-btn"
+      id="copyCodeButton"
```
Blackbaud-SteveBrush: Avoid using `id` since this could be duplicated several times on a page. You can target elements using Angular's Template Reference Variables.

```
<div stacheClipboard clipboardTrigger="copyCodeButton">
...
<button #copyCodeButton>
```

src/app/public/src/modules/clipboard/clipboard.directive.ts
```diff
@@ -0,0 +1,80 @@
...
+  selector: '[stacheClipboard]'
+})
+export class StacheClipboardDirective implements AfterViewInit, OnDestroy {
+  // --------------------------------------------------------------
+  // NOTE: Only supported for Code-Block component as of July 2018
+  // --------------------------------------------------------------
+
+  private defaultTigger = 'clipboardButton';
+
+  @Input() public clipboardTrigger: string = this.defaultTigger;
```
Blackbaud-SteveBrush: I added a comment about Template Reference Variables below, but this should be a `ElementRef` property.

See: https://angular.io/guide/template-syntax#ref-vars

src/app/public/src/modules/clipboard/clipboard.directive.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  public ngAfterViewInit() {
+    this.targetElm = this.el.nativeElement;
+    this.instantiateButton();
+    this.createEventListeners();
+    this.cdr.detectChanges();
+  }
+
+  public ngOnDestroy() {
+    this.clipboardSrv.destroy();
```
Blackbaud-SteveBrush: Angular services can implement `ngOnDestroy`, so this would be redundant. Make sure you add `StacheClipboardService` to this directive's providers!

See: https://stackoverflow.com/questions/45898948/angular-4-ngondestroy-in-service-destroy-observable

src/app/public/src/modules/clipboard/clipboard.directive.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (succeeded) {
+      this.cbOnSuccess.emit({ isSuccess: true, content: copiedContent });
+    } else {
+      this.cbOnError.emit({ isSuccess: false });
+    }
+  }
+
+  private createEventListeners(): void {
+    if (this.button) {
+      this.renderer.listen(this.button, 'click', () => {
```
Blackbaud-SteveBrush: This method reveals an implementation that's a bit muddy. You could instead add the directive to the trigger element (in this case, a button), and handle all the click events naturally. The target reference could also be passed in, so that the button would know what contents to copy.

Something like this?

```
<button
  stacheClipboard
  stacheClipboardTarget="stacheCodeBlock"
>
  Copy
</button>

<pre
  class="stache-code-output"
  stacheCodeBlock
>
  <code [ngClass]="getClassName()" [innerHTML]="output"></code>
</pre>
```

src/app/public/src/modules/clipboard/clipboard.service.ts
```diff
@@ -0,0 +1,89 @@
+// adapted from ngx-clipboard
+
+import { Inject, Injectable, Renderer2 } from '@angular/core';
+import { DOCUMENT } from '@angular/common';
+import { StacheWindowRef } from '../../..';
+
+@Injectable()
+export class StacheClipboardService {
+  private tempTextArea: HTMLTextAreaElement | undefined;
```
Blackbaud-SteveBrush: `undefined` is assumed; not needed?

src/app/public/src/modules/clipboard/clipboard.service.ts
```diff
@@ -0,0 +1,89 @@
...
+  constructor(
+    @Inject(DOCUMENT) private document: any,
+    private windowRef: StacheWindowRef
+  ) { }
+
+  /**
+   * copyFromInputElement
+   */
+  public copyFromInputElement(targetElm: HTMLInputElement | HTMLTextAreaElement, renderer: Renderer2): boolean {
+    try {
```
Blackbaud-SteveBrush: Avoid `try/catch` and only check known conditions. The reason is that `try/catch` could potentially handle an error you didn't intend for--such as an invalid type--(and your unit test should have caught to begin with). Instead, try to determine what the potential error could be, and let unknown errors bubble out naturally.

See: https://www.dotnetperls.com/tester-doer

src/app/public/src/modules/clipboard/clipboard.service.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (!this.tempTextArea) {
+      this.tempTextArea = this.createTempTextArea(this.document, this.windowRef);
+      this.document.body.appendChild(this.tempTextArea);
+    }
+    this.tempTextArea.value = content;
+    return this.copyFromInputElement(this.tempTextArea, renderer);
+  }
+
+  // remove temporary textarea if any
+  public destroy() {
```
Blackbaud-SteveBrush: See previous comment about services implementing `ngOnDestroy`.

src/app/public/src/modules/clipboard/clipboard.directive.ts
Blackbaud-TimPepper: Definitely. Some of these pieces are left overs from the implementation we are pulling the directive and directive service from that we should remove if we don't need/want them.

src/app/public/src/modules/code-block/code-block.component.ts
Blackbaud-TimPepper: That was the design I initially went with, but Brandon and I decided it would be more efficient to just have the icon update. This removes the need to add translation text and also gives the button a smaller footprint on the page, since there's no popover when you click. I'd be happy to discuss this further if you have additional thoughts.

src/app/public/src/modules/code-block/code-block.component.html
```diff
@@ -1,6 +1,17 @@
...
-    [innerHTML]="output"></code></pre>
-</div>
\ No newline at end of file
+  <div #codeFromContent class="stache-code-input">
+    <ng-content></ng-content>
+  </div>
+  <div stacheClipboard [attr.clipboardTriggerButton]="copyCodeButton">
+    <pre class="stache-code-output"><code [ngClass]="getClassName()" [innerHTML]="output"></code></pre>
+    <button
+      aria-label="copy to clipboard"
```
Blackbaud-BrandonJones: We are going to want to add this label through the SkyAppResources pipe, and store the value in the `resources_en_US.json` file.  For accessibility we may end up having to support several translations for these fields.

Blackbaud-BobbyEarl:
@Blackbaud-TimPepper  This is all non-blocking feedback:

- Have you/CDD given any thought to visually styling the code blocks a little different.  I'm really a fan of a couple choices the MS docs have made ([example](https://docs.microsoft.com/en-us/aspnet/core/razor-pages/?view=aspnetcore-2.1&tabs=visual-studio)), where they list the `languageType` and an always visible copy button in a "header" above the code block.

- While the `execCommand('copy')` seems really widely available, would it be worth wrapping in a `document.queryCommandSupported('copy')` to determine if the component is shown?

- I like the status changing to a check if it's copied, but could you make that check based on the return value of the `execCommand`?  https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand . Related, would it be worth changing the text to past tense during that state, so "copied"?

Blackbaud-TimPepper:
We haven't talked about re-styling at all, but I think that the Microsoft implementation looks really good. It would probably be helpful for our users as well, since it does clearly state which language is being used. 

I agree that doing a check to see if the command has fired is a good idea, and would like to look in to it. We were avoiding going with "copied" due to translation issues, but since we have moved past that I think that having copied in there on a successful copy would be useful.

Blackbaud-BrandonJones:
We should definitely create a different story to consider restyling the codeblock to feature the language type, probably wrap it up with our story to expand supported languages.

I don't think switching to the past tense of `Copied` is necessary as we have the visual feedback of the check to display it's been copied.  This avoids translation issues, as well as having to add a second input for user input text.

Agreed on the wrapping check for supported commands, and adding the check based on the return value of that exec command.

src/app/public/src/modules/clipboard/clipboard.service.ts
```diff
@@ -0,0 +1,75 @@
...
+    this.renderer.selectRootElement(inputElement).select();
+    this.renderer.selectRootElement(inputElement).setSelectionRange(0, inputElement.value.length);
+  }
+
+  private clearSelection(inputElement: HTMLInputElement | HTMLTextAreaElement) {
+    inputElement.blur();
+    this.windowRef.nativeWindow.getSelection().removeAllRanges();
+  }
+
+  private createTextAreaFromElement(element: HTMLElement): HTMLTextAreaElement {
```
Blackbaud-BrandonJones: @Blackbaud-SteveBrush left comments on the other PR:
In src/app/public/src/modules/clipboard/clipboard.service.ts:

> @@ -43,18 +51,25 @@ export class StacheClipboardService {
     this.windowRef.nativeWindow.getSelection().removeAllRanges();
   }
 
-  private createCopyTextArea() {
-    const copyTextAreaStyles = `
-      font-size: 12px;
-      border: 0;
-      padding: 0;
-      margin: 0;
-      position: absolute;
-      left: -9999px;
-      top: ${this.windowRef.nativeWindow.pageYOffset || this.document.documentElement.scrollTop}px;`;
+  private createTextAreaFromElement(element: HTMLElement): HTMLTextAreaElement {
+    if (!this.copyTextArea) {
Part of me is wondering if this HTML/CSS "stuff" could be added to the clipboard template instead. It just feels wrong to be writing HTML in the component class.

clipboard.component.html

(The variable names are inconsequential, but something more manageable like the following...)

<ng-template [ngIf]="copiedContents">
  <textarea
    class="stache-clipboard-hidden-contents"
    readonly
  >{{ copiedContents }}</textarea>
</ng-template>

Blackbaud-TimPepper: This text area is being generated by the Service to create something copy-able out of html content that isn't a text input/area. I think removing it from the Service code itself might cripple the the ability for the service to be reused later, outside of our implementation. If we are only ever going to use the service with the component itself, then I see no issue with moving it in to the component code and grabbing it programmatically as needed

Blackbaud-BrandonJones: This also prevents a new `textarea` from spawning per component.

Blackbaud-SteveBrush: Good points.

src/app/public/src/modules/clipboard/clipboard.component.ts
```diff
@@ -0,0 +1,31 @@
...
+  @Input()
+  public copyTarget: HTMLElement;
+
+  @Input()
+  public buttonText: string;
+
+  public buttonActive: boolean = false;
+
+  constructor(
+    private clipboardService: StacheCopyToClipboardService
```
Blackbaud-SteveBrush: @Blackbaud-TimPepper @Blackbaud-BrandonJones I concur with Bobby's request to change the button's text to "Copied!" after a successful operation.

This can be done easily with the `SkyAppResourcesService`.

**clipboard.component.ts**

```
this.resourcesService.getString('stache_clipboard_copied_button').subscribe((value) => {
  this.buttonText = value;
});
```

src/app/public/src/modules/clipboard/clipboard.service.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  public copyContent(element: HTMLElement) {
+    let copyTarget = this.getValidInputElement(element);
+    this.copyFromInputElement(copyTarget);
+  }
+
+  private copyFromInputElement(copyTarget: HTMLInputElement | HTMLTextAreaElement) {
+    this.selectTarget(copyTarget);
+    this.document.execCommand('copy');
```
Blackbaud-SteveBrush: I believe Bobby mentioned this already, but returning the success/failure of `execCommand` would be helpful for showing status markers on the button.

src/app/public/src/modules/code-block/code-block.component.html
```diff
@@ -1,6 +1,14 @@
...
+    <ng-content></ng-content>
+  </div>
+  <div>
+    <pre #stacheCodeOutput class="stache-code-output"><code
+        [ngClass]="getClassName()"
+        [innerHTML]="output"></code></pre>
+  </div>
+  <stache-copy-to-clipboard
+    [copyTarget]="stacheCodeOutput"
+    buttonText="{{ 'stache_copy_to_clipboard' | skyAppResources }}">
```
Blackbaud-SteveBrush: Instead of setting this here, we should probably make it the default within the clipboard component itself (I noticed that the clipboard doesn't set a default for the button text).

**clipboard.component.ts**

```
@Input()
public set buttonText(value: string) {
  this._buttonText = value;
}

public get buttonText(): string {
  return this._buttonText || this.defaultButtonText;
}

private _buttonText: string;
private defaultButtonText: string;

constructor(
  private resourcesService: SkyAppResourcesService
) { }

public ngOnInit(): void {
  this.resourcesService.getString('stache_copy_to_clipboard')
    .subscribe((value: string) => {
      this.defaultButtonText = value;
    });
}
```

src/app/public/src/modules/code-block/code-block.component.scss
```diff
@@ -13,7 +13,31 @@
   margin-bottom: $stache-margin * 2;
 }
 
-/deep/
+.stache-code-block {
+  position: relative;
+
+  ::ng-deep
+  .stache-clipboard-btn {
```
Blackbaud-SteveBrush: More of a question: should we put these styles in the clipboard style sheet to maintain consistency for when it's used outside of the code block?

src/app/public/src/modules/code-block/code-block.component.ts
```diff
@@ -43,13 +43,18 @@ export class StacheCodeBlockComponent implements AfterViewInit {
       this._languageType = this.defaultLanguage;
     }
   }
+
   public get languageType(): string {
     return this._languageType;
   }
 
+  @ViewChild('stacheCodeOutput')
+  public outputElement: any;
```
Blackbaud-SteveBrush: I don't believe this is needed anymore?

src/app/public/src/modules/code-block/code-block.component.ts
```diff
@@ -43,13 +43,18 @@ export class StacheCodeBlockComponent implements AfterViewInit {
...
+
   @ViewChild('codeFromContent')
   public codeTemplateRef: any;
 
+  public clipboardButtonClicked: boolean = false;
```
Blackbaud-SteveBrush: I don't believe this is needed anymore?

src/assets/locales/resources_en_US.json
```diff
@@ -11,6 +11,14 @@
     "message": "Toggle sidebar",
     "_description": "Button text to show or hide sidebar menu"
   },
+  "stache_copy_to_clipboard_button_title": {
+    "message": "Copy to clipboard",
+    "_description": "Button title to for the copy to clipboard button"
+  },
+  "stache_copy_to_clipboard": {
```
Blackbaud-SteveBrush: I would suffix this with `_button`. Right now it's too generic and isn't immediately obvious (when viewed in the templates) what its purpose is.

src/app/public/src/modules/code-block/code-block.component.scss
Blackbaud-BrandonJones: These styles are very particular to the code-block itself.  The styles of the button itself are just those of the sky button.  I think consuming components should be able to position the button however it is ideal for that component.  
e.g  a form input wouldn't want to have the button in the same place as the code block does.

src/app/public/src/modules/code-block/code-block.component.html
Blackbaud-BrandonJones: I debated with this.  The reason it's not setting the default text within with component is to avoid forcing the button to contain text at all.  If the user just wanted the icon (for example next to a form Input) they don't have to supply text.

src/app/public/src/modules/clipboard/clipboard.component.ts
Blackbaud-BrandonJones: This would also require us to support a second user input for `copiedText` from the user.  Also the [microsoft docs](https://docs.microsoft.com/en-us/aspnet/core/razor-pages/?view=aspnetcore-2.1&tabs=visual-studio) don't change the value either.  Not sure it's necessary?

src/app/public/src/modules/clipboard/clipboard.service.ts
```diff
@@ -0,0 +1,79 @@
...
+    this.document = this.windowRef.nativeWindow.document;
+  }
+
+  public copyContent(element: HTMLElement) {
+    let copyTarget = this.getValidInputElement(element);
+    this.copyFromInputElement(copyTarget);
+  }
+
+  public verifyCopyCommandEnabled() {
+    return this.document.execCommand('copy');
```
Blackbaud-BrandonJones: We should use `queryCommandSupported`, if you load a page before any user interaction, execCommand returns false.

Blackbaud-BrandonJones:
I think this looks good to me, @Blackbaud-SteveBrush @Blackbaud-BobbyEarl ?

PR #500, opened 2018-07-09
Added z index to back to top button

Blackbaud-BrandonJones:
Added a z-index to the back to top button.  This is to prevent it from being overlapped by the footer.


Blackbaud-BobbyEarl:
Adjusting the `z-index` is definitely the quickest fix.  Long term, should the `back-to-top` component even be on top of the footer?  Meaning, should there be similar logic used in the sidebar?

Blackbaud-BrandonJones:
I think the button over lapping the footer in this instance is fine.  There is no information that gets displayed on that side of the browser window, and adding the similar calculations to that button will be very taxing on older browsers.

PR #506, opened 2018-07-12
Clipboard component

Blackbaud-BrandonJones:


src/app/public/src/modules/clipboard/clipboard.service.ts
```diff
@@ -43,18 +51,25 @@ export class StacheClipboardService {
...
-    const copyTextAreaStyles = `
-      font-size: 12px;
-      border: 0;
-      padding: 0;
-      margin: 0;
-      position: absolute;
-      left: -9999px;
-      top: ${this.windowRef.nativeWindow.pageYOffset || this.document.documentElement.scrollTop}px;`;
+  private createTextAreaFromElement(element: HTMLElement): HTMLTextAreaElement {
+    if (!this.copyTextArea) {
```
Blackbaud-SteveBrush: Part of me is wondering if this HTML/CSS "stuff" could be added to the clipboard template instead. It just feels wrong to be writing HTML in the component class.

**clipboard.component.html**

(The variable names are inconsequential, but something more manageable like the following...)

```
<ng-template [ngIf]="copiedContents">
  <textarea
    class="stache-clipboard-hidden-contents"
    readonly
  >{{ copiedContents }}</textarea>
</ng-template>
```

src/app/public/src/modules/clipboard/clipboard.service.ts
```diff
@@ -43,18 +51,25 @@ export class StacheClipboardService {
...
+        const copyTextAreaStyles = `
+        font-size: 12px;
+        border: 0;
+        padding: 0;
+        margin: 0;
+        position: absolute;
+        left: -9999px;
+        top: ${this.windowRef.nativeWindow.pageYOffset || this.document.documentElement.scrollTop}px;`;
```
Blackbaud-SteveBrush: What's the reasoning behind calculating the `top` value, and not just setting it to something negative? `-100000px` or something?

PR #512, opened 2018-07-19
Feature table of contents rework

Blackbaud-TimPepper:
Rework the Table of Contents component to sit on the right sidebar and track the user down the page as they scroll/navigate/click. 

The table of contents disappears as the screen gets skinnier. I did not add the ability to toggle it like the sidebar, as in page nav is less important than site wide nav.


Blackbaud-SteveBrush:
@Blackbaud-TimPepper I see a large gap on the right-hand side of some pages. Is that intended?

screen shot 2018-07-20 at 10 35 35 am

src/app/public/src/modules/table-of-contents/table-of-contents-wrapper.component.scss
```diff
@@ -0,0 +1,26 @@
...
+
+  .stache-table-of-contents-content {
+    width: $stache-sidebar-column-width;
+    padding-top: $stache-padding-half;
+    padding-bottom: $stache-padding-plus-half;
+    padding-right: $stache-padding-half;
+    max-height: 100%;
+
+    .stache-affix-top {
+      overflow-y: scroll;
```
Blackbaud-BrandonJones: This would need to be `overflow-y: auto`

Blackbaud-BrandonJones: Also should this be on the affix top? or on the `table-of-contents-content`

Blackbaud-TimPepper: I went ahead and tested it and it looks like it must be in the `stache-affix-top` element for the scrolling to trigger.

Blackbaud-BrandonJones:
This seems to be having issues with the nested anchors in the demos/anchor-links pages.  They don't seem to register as active in the TOC when scrolling.

src/app/public/src/modules/table-of-contents/table-of-contents.component.ts
```diff
@@ -13,9 +22,76 @@ export class StacheTableOfContentsComponent implements StacheNav, AfterViewInit
...
+    this.routes.map((route: StacheNavLink) => {
+      if (route.offsetTop <= this.pageOffset) {
+        this.activeRoute = route;
+      }
+    });
+  }
+
+  // Updates ng-class level trigger to show/hide blue border marking location on page
+  private updateView() {
+    this.routes.map((route: StacheNavLink) => {
```
Blackbaud-BrandonJones: Using map is pretty expensive here, as it creates a new array for each time it's run.  Here we are running it on every trigger of scroll.  Can we just update the existing route/array?

Blackbaud-BrandonJones: maybe just use `this.routes.forEach`?

src/app/public/src/modules/nav/nav.component.html
```diff
@@ -3,7 +3,7 @@
     <li class="stache-nav-list-item"
       *ngFor="let route of routes"
       [attr.aria-current]="route.isCurrent ? 'page' : false"
-      [ngClass]="{'active': route.isActive, 'current': route.isCurrent}">
+      [ngClass]="{'active': route.isActive, 'current': route.isCurrent, 'stache-table-of-contents-active': route.isActiveTocAnchor}">
```
Blackbaud-BrandonJones: Can we just make use of one of these already existing classes instead of adding a new one?  `isCurrent` would make sense in this situation?

src/app/public/src/modules/table-of-contents/table-of-contents.component.ts
Blackbaud-BrandonJones: We can also make use of the parameters in the `forEach` call to check other elements in the routes.   Something like 
```
  private updateView() {
    this.routes.forEach((route: StacheNavLink, index: number) => {
      if (this.isActiveRoute(route, index)) {
        route.isCurrent = true;
        this.activeRoute = route;
      } else {
        route.isCurrent = false;
      }
    });
  }

  private isActiveRoute(route: StacheNavLink, index: number): boolean {
    if ((this.window.innerHeight + 5) >= this.documentBottom) {
      this.activeRoute = this.routes[this.routes.length - 1];
      return route === this.activeRoute;
    }

    return route.offsetTop <= this.pageOffset
      && (this.routes[index + 1] === undefined || this.routes[index + 1].offsetTop > this.pageOffset);
  }
```
Would allow us to eliminate the updateActiveRoute method, and make it one `forEach` loop instead of two.

src/app/public/src/modules/table-of-contents/table-of-contents.component.ts
```diff
@@ -13,9 +22,76 @@ export class StacheTableOfContentsComponent implements StacheNav, AfterViewInit
...
+        route.isActiveTocAnchor = false;
+      }
+    });
+  }
+
+  // Checks to see if stache-wrapper exists and provides offset from top of window
+  private getContentOffset() {
+    this.headerHeight =
+      this.documentElement.querySelector('.stache-wrapper') &&
+      this.documentElement.querySelector('.stache-wrapper').offsetTop;
```
Blackbaud-BrandonJones: Not sure I fully understand this?  We are assigning the HeaderHeight to a boolean here.  But we are using it as a number elsewhere?

src/app/public/src/modules/table-of-contents/table-of-contents.component.scss
```diff
@@ -28,11 +31,11 @@
...
-  padding: 6px 0 12px 19px;
+  padding: 6px 0;
 }
 
 .stache-table-of-contents-heading {
-  margin: 0 0 .7em;
-  font-size: 1.1em;
+  font-size: 1.5em;
+  font-weight: 300;
+  border-left: 8px solid transparent;
```
Blackbaud-BrandonJones: Can we achieve the same affect by using padding or margin?  Seems strange to use a border transparent to adjust a block element.

Blackbaud-BrandonJones: We use it in the Sidebar because the title can also be the `active` route, so it sometimes get filled with color.  Here it wouldn't make sense.

src/app/public/src/modules/table-of-contents/table-of-contents.component.ts
Blackbaud-TimPepper: This is basically checking to see if the element exists before pulling the offsetTop value off of it. It returns a number.

src/app/public/src/modules/table-of-contents/table-of-contents.component.scss
Blackbaud-TimPepper: If I remember correctly I was using this to assist in padding so when the blue line/border didn't exist it wasn't disappearing, it was just becoming transparent. This makes it so that the text is always a consistent distance away from the blue indicator.

Blackbaud-BrandonJones: Correct, but in this case, we never set the `Content` header to have a border, so it won't shift.

src/app/public/src/modules/table-of-contents/table-of-contents.component.ts
Blackbaud-BrandonJones: Very strange... looks like an expression?  If we set the  element to a variable we only have to query and search the dom once.  
```
  private headerHeight: number = 0;

private getContentOffset() {
    let wrapperEl = this.documentElement.querySelector('.stache-wrapper');

    if (wrapperEl) {
      this.headerHeight = wrapperEl.offsetTop;
    }
  }
```

Blackbaud-BobbyEarl:
I like the "Contents" section still being visible on smaller resolutions, just inline at the top of the document vs a dedicated column.

Issue #507, opened 2018-07-12
Codeblock Style update


Blackbaud-BrandonJones:
We should consider adding a "header" to the Stache code blocks, it could display the `languageType` and also contain the copy button.  See MS docs for the example 
https://docs.microsoft.com/en-us/aspnet/core/razor-pages/?view=aspnetcore-2.1&tabs=visual-studio

Blackbaud-TimPepper:
This has been resolved by [#515](https://github.com/blackbaud/stache2/pull/515)

A header has been added that displays the code language and copy-to-clipboard button

PR #515, opened 2018-07-26
add code-block header

Blackbaud-TimPepper:
Show what language is being displayed in the code-block and move the copy-to-clipboard button in to the header so it is always visible. 

https://github.com/blackbaud/stache2/issues/507

### DIFF
--- a/src/app/demos/code-block/index.html
+++ b/src/app/demos/code-block/index.html
@@ -3,13 +3,6 @@
   showBreadcrumbs="false"
   showEditButton="true">
 
-<stache-code-block
-  languageType="js">
-  $(document).ready(() => {
-    // do something
-  });
-</stache-code-block>
-
   <stache-code-block
     languageType="js">
     $(document).ready(() => {
--- a/src/app/demos/code-block/index.html
+++ b/src/app/demos/code-block/index.html
@@ -3,6 +3,13 @@
   showBreadcrumbs="false"
   showEditButton="true">
 
+<stache-code-block
+  languageType="js">
+  $(document).ready(() => {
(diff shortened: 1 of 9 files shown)


Blackbaud-TimPepper:
image

Blackbaud-TimPepper:
Working design with feedback from Todd:
image

Blackbaud-SteveBrush:
@Blackbaud-ToddRoberts approved this design?

Blackbaud-TimPepper:
Yep! Todd actually suggested the final design changes I made. And I already confirmed that the new design works with the tabbed implementation.

Blackbaud-BobbyEarl:
I realize this behavior is not changing in your PR, but it stands out to me more now that we're visually displaying the language type in the header.  Passing either an unknown `languageType` or not setting the property at all displays "markup" in the header.  Just bringing it up.

The styling for the header seems a little off.  The font-size is 19px and because of the margin around the button, it seems bulkier than it needs to be.  Would it be reduce those to `15px` and have the same spacing as buttons in a toolbar?

I realize this isn't an approved use of a "toolbar," but it seems we'd still want to emulate the style in this case.  https://developer.blackbaud.com/skyux/components/toolbar

Blackbaud-ToddRoberts:
My suggestion was to use sky-subsection-heading for the language text. If the language is not that important to warrant the larger size, and/or largely self-evident with a more subtle reminder appropriate, we could use sky-deemphasized text there.

Blackbaud-TimPepper:
Updated view. Tried to post a screenshot for you guys, but github is refusing to cooperate.

Issue #527, opened 2018-09-06
Add a "caption" input to images


Blackbaud-ThomasOrtiz:
### Expected behavior
I should be able to add an optional `Description`  that displays centered underneath the image.

![image](https://user-images.githubusercontent.com/35976987/45178676-59ad7580-b1e4-11e8-9b71-a74fb8e2c1ec.png)

### Actual behavior
This is not currently so we can not give any additional context to images.

Blackbaud-BrandonJones:
Resolved in release 2.11.4
https://github.com/blackbaud/stache2/pull/530

PR #528, opened 2018-09-06
#527 - Add "caption" input to image component

Blackbaud-ThomasOrtiz:
Address #527 

## Additions
* Added an optional `caption` input to the `stache-image` component.
* Added a `stache-image` demo page.

### DIFF
--- a/src/app/demos/image/index.html
+++ b/src/app/demos/image/index.html
@@ -1,14 +1,26 @@
-<stache pageTitle="Image" show showBreadcrumbs="false" showEditButton="true" showTableOfContents="true">
+<stache pageTitle="Image" showBreadcrumbs="false" showEditButton="true" showTableOfContents="true">
 
   <stache-page-anchor>Simple image</stache-page-anchor>
   <div style="width: 400px;">
     <stache-image imageSource="~/assets/hero-sample-1.jpg"></stache-image>
   </div>
 
-
-  <stache-page-anchor>Image with a description</stache-page-anchor>
+  <stache-page-anchor>Image with a short caption</stache-page-anchor>
   <div style="width: 400px;">
-    <stache-image imageSource="~/assets/hero-sample-1.jpg" description="Figure 1: A hero image with a description."></stache-image>
+    <stache-image imageSource="~/assets/hero-sample-1.jpg" caption="A"></stache-image>
+  </div>
+
+  <stache-page-anchor>Image with an average caption</stache-page-anchor>
+  <div style="width: 400px;">
+    <stache-image imageSource="~/assets/hero-sample-1.jpg" caption="Figure 1: A hero image with a caption."></stache-image>
+  </div>
+
+  <stache-page-anchor>Image with a long caption</stache-page-anchor>
+  <div style="width: 400px;">
+    <stache-image
+      imageSource="~/assets/hero-sample-1.jpg"
+      caption="Figure 2: A hero image with a caption. Such a long caption in fact that it may go to a new line. Its crazy.">
+    </stache-image>
   </div>
 
 </stache>
--- a/src/app/demos/image/index.html
+++ b/src/app/demos/image/index.html
@@ -1,0 +1,14 @@
+<stache pageTitle="Image" show showBreadcrumbs="false" showEditButton="true" showTableOfContents="true">
+
+  <stache-page-anchor>Simple image</stache-page-anchor>
+  <div style="width: 400px;">
+    <stache-image imageSource="~/assets/hero-sample-1.jpg"></stache-image>
+  </div>
+
+
+  <stache-page-anchor>Image with a description</stache-page-anchor>
+  <div style="width: 400px;">
+    <stache-image imageSource="~/assets/hero-sample-1.jpg" description="Figure 1: A hero image with a description."></stache-image>
+  </div>
+
+</stache>
--- a/src/app/demos/image/index.html
+++ b/src/app/demos/image/index.html
@@ -19,7 +19,7 @@
   <div style="width: 400px;">
     <stache-image
       imageSource="~/assets/hero-sample-1.jpg"
-      caption="Figure 2: A hero image with a caption. Such a long caption in fact that it may go to a new line. Its crazy.">
+      caption="Figure 2: A hero image with a caption. Such a long caption, in fact, that it may go to a new line. Its crazy.">
     </stache-image>
   </div>
 
--- a/src/app/public/src/modules/image/image.component.html
+++ b/src/app/public/src/modules/image/image.component.html
@@ -1,7 +1,7 @@
-<div class="stache-image"><div class="stache-image"></div>
+<div class="stache-image">
   <img class="stache-image-object" [src]="imageSource" [alt]="imageAlt" />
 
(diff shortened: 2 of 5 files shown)


src/app/public/src/modules/image/image.component.html
```diff
@@ -1,3 +1,7 @@
 <div class="stache-image">
   <img class="stache-image-object" [src]="imageSource" [alt]="imageAlt" />
+
+  <div *ngIf="description" class="stache-image-description">
+      <span><i>{{ description }}</i><span>
```
Blackbaud-TimPepper: Invalid closing tag for span here, causing build to fail

src/app/public/src/modules/image/image.component.spec.ts
```diff
@@ -34,4 +34,10 @@ describe('StacheImageComponent', () => {
     expect(component.imageSource).toBe(imageSource);
   });
 
+  it('should have a description input', () => {
+    component.description = this.description;
```
Blackbaud-TimPepper: We should define `description` for readabilities sake

src/app/demos/image/index.html
```diff
@@ -0,0 +1,14 @@
+<stache pageTitle="Image" show showBreadcrumbs="false" showEditButton="true" showTableOfContents="true">
```
Blackbaud-BrandonJones: I think you have an errant `show` directive here that needs to be removed.

src/app/public/src/modules/image/image.component.ts
```diff
@@ -11,4 +11,7 @@ export class StacheImageComponent {
 
   @Input()
   public imageAlt: string = '';
+
+  @Input()
+  public description: string = '';
```
Blackbaud-BrandonJones: Not sure description is the right label we should go with, that sounds more like something that would be added to an `alt` or `title` attribute on an image.  Perhaps something like `caption`?

Blackbaud-BrandonJones: I would also leave this value as `undefined` if not specifically defined by a user. Since this is a conditional field, having a default value means this is defined.

src/app/public/src/modules/image/image.component.html
Blackbaud-BrandonJones: I am not `100%` on the use of the `<i>` tags here, if they are appropriate.  Maybe @Blackbaud-ToddRoberts has more feedback, but I feel like we would be better served using a [SKY UX](https://developer.blackbaud.com/skyux/design/styles/typography) class for the span.  Maybe `sky-field-label` or `sky-deemphasized`?  Other thoughts about the text being left aligned vs center.

src/app/demos/image/index.html
```diff
@@ -0,0 +1,14 @@
...
+
+  <stache-page-anchor>Simple image</stache-page-anchor>
+  <div style="width: 400px;">
+    <stache-image imageSource="~/assets/hero-sample-1.jpg"></stache-image>
+  </div>
+
+
+  <stache-page-anchor>Image with a description</stache-page-anchor>
+  <div style="width: 400px;">
+    <stache-image imageSource="~/assets/hero-sample-1.jpg" description="Figure 1: A hero image with a description."></stache-image>
```
Blackbaud-BrandonJones: Thanks for updating the Demo to include this!  Once we square away the updates, can we also get an example of How this looks with Really long and really short descriptions?

src/app/public/src/modules/image/image.component.spec.ts
Blackbaud-BrandonJones: Also a test to be sure the value is undefined should the description not be set by a user would be valued as it is conditional.

Blackbaud-BrandonJones:
Thanks for the PR and issue, few comments and updates requested, love the ambition and contributions here.

src/app/public/src/modules/image/image.component.html
Blackbaud-ToddRoberts: Assuming this is a caption for an image? If so I'd recommend `sky-deemphasized` class rather than `<i>`, and left justify text.

src/app/public/src/modules/image/image.component.ts
Blackbaud-ThomasOrtiz: I 100% agree with `caption`. When looking at this I knew there was a better word for it but just couldn't think of it haha.

src/app/public/src/modules/image/image.component.html
Blackbaud-ThomasOrtiz: I originally looked at the typography page that you linked but didn't find a class that seemed to fit really well based off the description. If I had to pick between those two I'd say `sky-deemphasized`.

I think centering will create a cleaner look -- here's some examples:

image

Blackbaud-ThomasOrtiz:
@Blackbaud-BrandonJones @Blackbaud-TimPepper 
I have updated all of the changes you outlines above.

@Blackbaud-ToddRoberts 
As far as left aligned versus centered goes, I think centering will create a cleaner look -- here's some examples:

image

Blackbaud-ThomasOrtiz:
@Blackbaud-ToddRoberts Here is an example of the caption left aligned:

image

Blackbaud-ToddRoberts:
Left justified is easier to read for multi-line captions and easier to find for very short captions, and seems to be standard if you look at e.g. major newspapers.

Blackbaud-BrandonJones:
I agree with Todd. the left align is easier to follow with your eye as you go down the page as well from the image.  It creates a clean line and direction.

src/app/public/src/modules/image/image.component.html
```diff
@@ -1,3 +1,7 @@
 <div class="stache-image">
   <img class="stache-image-object" [src]="imageSource" [alt]="imageAlt" />
+
+  <div *ngIf="caption" class="stache-image-caption">
+      <span class="sky-deemphasized">{{ caption }}</span>
```
Blackbaud-BrandonJones: Your indent here needs adjusting,  2 spaces.

src/app/public/src/modules/image/image.component.scss
```diff
@@ -10,3 +10,8 @@
   max-width: 100%;
   height: auto;
 }
+
+.stache-image-caption {
+  margin: 0.5em 1em 0.5em 1em;
+  text-align: center;
```
Blackbaud-BrandonJones: Moving to left align, we can us padding rather than margin here.  `padding: 0.5em 0;`

Blackbaud-BrandonJones:
Thank you so much for the contribution! We will try to get a release out today to make this available.

Blackbaud-ThomasOrtiz:
Yup, no problem 👍 

Thanks for the quick and useful feedback! This was a really easy feature to get out with all of us involved.

Blackbaud-ThomasOrtiz:
@Blackbaud-BrandonJones Also, do I manually delete my branch or is that something that you all will handle?

Blackbaud-BrandonJones:
Your branch is in your own fork :) So that's up to you to manage.

Blackbaud-ThomasOrtiz:
Oh thats a "duh moment" for me haha. And I imagine you all will close #527 once its deployed or should we close it now that its merged into master?

PR #531, opened 2018-09-14
Hide navbar links

Blackbaud-TimPepper:


src/app/demos/route-metadata/route-order-test/a/index.html
```diff
@@ -1,5 +1,6 @@
 <stache
   pageTitle="a"
-  navOrder="9999">
+  navOrder="9999"
+  hideFromSidebar="true">
```
Blackbaud-BrandonJones: As we will be extending this flag to also keep the pages from being published to the search service or discoverable at all, would calling this something like `publish: false` or `unpublished: true`?  Something along those lines that works for more than just hiding it from the sidebar?

Blackbaud-LindseyRix: in Stache1 we had two options: 
- showInNav: true or false (because you may want it published but not in navigation)
- published: true or false (this enables you to have a page that you want to include in your PR, but not be live yet). Branching may be better for `published`, but sometimes it could be easier for multiple people to work on the same unpublished page and not in a branch.

Blackbaud-BrandonJones: Good points, I'd agree with calling it `showInNav`, and for the time being encourage the use of branches for publishing.

src/app/public/src/modules/sidebar/sidebar.component.ts
```diff
@@ -34,12 +34,25 @@ export class StacheSidebarComponent implements StacheNav, OnInit {
...
+  }
+
+  private filterHiddenRoutes(routes: StacheNavLink[]) {
+    if (routes && routes.length) {
+      return routes.filter(route => {
+        route.children = this.filterHiddenRoutes(route.children);
+        if (route.showInNav !== false) {
+          return route;
+        }
+      }).filter(child => !!child);
```
Blackbaud-BrandonJones: Is there a case that we would need this extra filter?

Blackbaud-BrandonJones:
This will be approved once the required plugin PR is merged and released.  This PR can then be updated to point to that specific version in the dependencies.

src/app/home.component.ts
```diff
@@ -46,11 +46,13 @@ export class HomeComponent {
...
-          path: '/0'
+          path: '/0',
+          showInNav: true
         },
         {
-          name: 'Sidebar Link',
-          path: '/1'
+          name: 'Hidden Sidebar Link',
+          path: '/1',
+          showInNav: false
```
Blackbaud-BrandonJones: Don't think these changes are needed any longer.  Custom sidebar routes can achieve this by simply not adding the route to begin with.

src/app/demos/route-metadata/route-order-test/a/index.html
```diff
@@ -1,5 +1,6 @@
 <stache
   pageTitle="a"
-  navOrder="9999">
+  navOrder="9999"
+  showInNav="false">
```
Blackbaud-BrandonJones: Having a sample secret page I don't think we should hide other example routes.

src/app/public/src/modules/shared/route.service.ts
```diff
@@ -94,8 +94,8 @@ export class StacheRouteService {
         },
         pathMetadata);
 
-      return formattedRoute as any;
-    });
+        return formattedRoute as any;
```
Blackbaud-BrandonJones: extra spaces here, should align with the `let` statement.

src/app/home.component.ts
```diff
@@ -46,7 +46,8 @@ export class HomeComponent {
       children: [
         {
           name: 'Sidebar Link is super long and even longer foo bar baz',
-          path: '/0'
+          path: '/0',
+          showInNav: true
```
Blackbaud-BrandonJones: This one too and we are good to go :)

PR #532, opened 2018-09-17
update prism to latest and add all new languages

Blackbaud-TimPepper:


Blackbaud-BrandonJones:
Looks good, non blocking thought:  How would you feel about replacing `-` with a space when in the `setDisplayName` method?  This would allow for things like `visual-basic` to display as `visual basic`.

Issue #533, opened 2018-09-17
Edit button should open link in new tab/window


Blackbaud-ShaydeNofziger:
### Expected behavior

When clicking the edit button component's anchor tag, I expect a new tab to open to the repo's file for editing.

### Actual behavior

The repo's file is opened in the same window I am viewing the documentation in, making it more cumbersome/difficult to view the rendered page and code simultaneously.

Blackbaud-BrandonJones:
After mulling it over and discussing it, we are going to stick with the existing pattern we set for Stache sites.  Links open in the same window.  This allows users to decide for themselves which behavior they want the link to follow.  `right click + open in a new tab`,  `ctrl/cmd + click`, etc. for opening a new tab.
 
Thank you for the PR and filing an issue, we very much appreciate the efforts in contributing to Stache.

PR #534, opened 2018-09-17
Open edit page in new tab/window

Blackbaud-ShaydeNofziger:
Include `target="_blank"`  in the `edit-button` component's anchor tag to open the edit page in a new window/tab.

This is more convenient to end users - they can view the content and code simultaneously without having to move back/forward between pages, or having to remember to "open in new tab".

#533 - associated issue

### DIFF
--- a/src/app/public/src/modules/edit-button/edit-button.component.html
+++ b/src/app/public/src/modules/edit-button/edit-button.component.html
@@ -1,5 +1,5 @@
 <div class="stache-edit-button" *ngIf="url">
-  <a class="stache-btn-edit" [href]="url">
+  <a class="stache-btn-edit" [href]="url" target="_blank">
     <i class="fa fa-edit" aria-hidden="true"></i>
     {{ editButtonText }}
   </a>


src/app/public/src/modules/edit-button/edit-button.component.html
```diff
@@ -1,5 +1,5 @@
 <div class="stache-edit-button" *ngIf="url">
-  <a class="stache-btn-edit" [href]="url">
+  <a class="stache-btn-edit" [href]="url" target="_blank">
```
Blackbaud-BrandonJones: Our general rule of thumb is that all links should open within the same tab, this allows users to right click and open in a new tab without forcing the rule on everyone.  Lots of debate around this, however because that is our default mode, we do have the option of opening in a new tab, but need to add both the `fa fa-external-link` icon after the text here, and `rel="noopener noreferrer"` to links containing target="_blank" as a precaution against reverse tabnabbing

Blackbaud-BrandonJones:
After mulling it over and discussing it, we are going to stick with the existing pattern we set for Stache sites.  Links open in the same window.  This allows users to decide for themselves which behavior they want the link to follow.  `right click + open in a new tab`,  `ctrl/cmd + click`, etc. for opening a new tab.
 
Thank you for the PR and filing an issue, we very much appreciate the efforts in contributing to Stache.

PR #535, opened 2018-09-18
update header styles

Blackbaud-TimPepper:


Blackbaud-BrandonJones:
I think the only other header tag that will need an update is the `stache-sidebar-heading` will need the margin updated there.

Blackbaud-TimPepper:
I considered the stache-sidebar-heading as well, but we aren't overwriting any sky styles there correct?

Blackbaud-BrandonJones:
Correct, but it’s no longer inheriting the styles from the h3 styles.  It should.

PR #546, opened 2018-10-31
Feature skip to content

Blackbaud-BrandonJones:
For accessibility we are providing a way to keyboard/screen reader navigating users to skip to the page content without need of going through all nav items first.  This design is based on the ideas from [WebAIM](https://webaim.org/techniques/skipnav/), a hidden link for users, that will be displayed when selected from the keyboard or screen reader.  

Behavior:
Skip to main content link will not display on page load.
Skip to main content link is focusable from keyboard or screen reader before primary nav items.
When the link is focused the link will display on the page
When the link loses focus, it remains visible for a few seconds so users who tab quickly will be able to see the link exists and can tab back to it.

Looking for feedback on the design and implementation.


Blackbaud-TimPepper:
I think that, as with WebAIM the link should be at tabindex="0". Seems like a solid proof of concept though. The location of it seems good, though should we make it more visible? I didn't notice it the first time I tabbed through.

Blackbaud-BrandonJones:
makes sense to me, I can look into adding a fly out style item like they use on webaim

Blackbaud-BrandonJones:
@Blackbaud-TimPepper update the style

Blackbaud-TimPepper:
Cool I like this new dropdown style. Definitely much more visible. Is there any way we can make the dropdown the first tab-able point of entry for the user? I think this is less important, but I do feel like I have to search for it a bit when tabbing around. Also, can you navigate directly to this element from a screen reader menu or do you have to tab to it directly? 

Unrelated thought: Are the skyux buttons supposed to be outside of the tab order?

Blackbaud-MattGregg:
@Blackbaud-BrandonJones Looking good. When I tried it, the link didn't seem to appear as the first element one tabs into on the page. The point mentioned 'Skip to main content link is focusable from keyboard or screen reader before primary nav items.' above. It appears after I tab through the site navigation bar.

Maybe this is a demo site oddity, but this is not the behavior I would expect given the purpose of this type of technique....allowing someone to skip over repeated navigation on every page of a web site to get right to the page's content. When someone navigates to a new page, the focus shouldn't be in the page any more. When they hit tab, the skip link appears as the first focusable element at the top and then can take it or move into the site nav bar. 

Here's a couple of examples - 
https://www.ncwit.org
https://www.starbucks.com/ - has a few other links as well (which are less common and I'm unsure of value/haven't researched)

Blackbaud-BobbyEarl:
- Agree with @Blackbaud-MattGregg about making it the first element to focus when tabing.
- I'm personally not a fan of the animation that's applied.

Blackbaud-SteveBrush:
@Blackbaud-MattGregg @Blackbaud-BobbyEarl

I'm curious if we should add an anchor to either the Omnibar, or to Host's main HTML page that skips to the primary "app" element. Then, Stache can provide its own skip-to-content link, that will be isolated to its features. My worry with Stache adding the link to the top of the document is that it could potentially interfere with a future skip-to-content link added by Omnibar (this should definitely be a feature of Omnibar, yes?), or it could interfere with a SPA consumer's own custom skip-to-content link. 

My suggestion is that omnibar/host would provide their own mechanism to skip its menu items, and Stache would provide its own mechanism to skip the sidebar navigation.

Thoughts on that?

Blackbaud-MattGregg:
@Blackbaud-SteveBrush I can see that being a good solution as well given how this is all architected. I hadn't proposed a skip link for the omnibar b/c there were higher priority A11y issues and the omnibar nav is built without that many tab stops for the components there so didn't seem too tedious to need to skip over. 
Interestingly I just checked new bb.com and it has a version of this built in (though after the site nav bar). If it's added to the omnibar, would be good to have solution work there too (and avoid being redundant with multiple skip links at the top)

Blackbaud-BrandonJones:
So... @Blackbaud-MattGregg @Blackbaud-BobbyEarl @Blackbaud-SteveBrush Was the concensus here that I can go ahead with this solution here? And an issue would be created within Omni bar for the skip nav link there?   If so, @Blackbaud-BobbyEarl you do not like the animation here, can you expand more on that with some suggestion as to an alternative you might like more?

Blackbaud-BobbyEarl:
I would prefer no animation at all.

My vote would be not to proceed with this technique.  I under the inclusion of the Omnibar makes it more of a technical challenge.  I do think; however, this shouldn't stop us from providing the correct experience.

In an ideal world, I like Starbucks as the perfect example.
- Contains no animation.
- Provides multiple "Skip To..." links for the three main areas (Nav, Content, and Footer).
- "Skip To..." links are the first tabbable elements on the page.

Blackbaud-MattGregg:
I agree the top of the page solution is preferred and if that is the Omnibar solution, that would be the one to prioritize. I don't have a good sense of how valuable having a additional skip link within the page sidebar navigation is. Probably doesn't harm anything, but not sure it's needed. 

I would keep it simple and just have skip to content be the only link.

Blackbaud-BrandonJones:
Sidebar nav items can get very long, as well as omnibar nav items, I would think we should provide the ability to skip each individually to provide a more ideal experience.

Blackbaud-BrandonJones:
This is being addressed in a new library from SKY, a  new PR will be opened using the library once it is released.

Issue #538, opened 2018-09-20
Be able to use action buttons without icons


Blackbaud-LindseyRix:
We need ability to have action buttons for pages that don't require you to provide an icon. If I don't provide an icon now, it's not spaced well for the size of the action button. 

![pasted image at 2017_08_08 03_36 pm](https://user-images.githubusercontent.com/11530192/29101574-2ac6f4a6-7c81-11e7-8786-c6064c896f8d.png)

Blackbaud-LindseyRix:
Thanks to Tim, this issue has been resolved and you longer have to use an icon on every action button. However, we recommend that you either use an icon every action button or none, just to maintain a consistent look and feel on a navigation page.

PR #548, opened 2018-11-02
remove icon element entirely if icon is not provided

Blackbaud-TimPepper:
This removes the extra padding noted in [#538](https://github.com/blackbaud/stache2/issues/538)


src/app/public/src/modules/action-buttons/action-buttons.component.spec.ts
```diff
@@ -120,9 +146,4 @@ describe('StacheActionButtonsComponent', () => {
 
     expect(component.filteredRoutes.length).toBe(3);
   });
```
Blackbaud-BrandonJones: We will want to add the unit test for accessibility in here

Blackbaud-BrandonJones:
LGTM

PR #551, opened 2018-11-09
Accessibility code block

Blackbaud-TimPepper:
Removed the background color and added more accessible (higher color contrast) styles for the code blocks.


Blackbaud-BrandonJones:
Build seems to be failing?

src/app/public/src/modules/code-block/code-block.component.scss
```diff
@@ -8,7 +9,6 @@
   border: 1px solid $stache-border-color-neutral-medium;
   border-top: none;
   margin-top: 0;
-  background-color: $stache-background-color-selected;
```
Blackbaud-BrandonJones: We might need to add a `background-color: #fff` if that's the direction we want to go.  This will prevent the background from being transparent on some layouts and cause issues with visibility.  A crude example:
screen shot 2018-11-13 at 1 33 51 pm

src/app/public/src/modules/code-block/prism-a11y.scss
```diff
@@ -0,0 +1,131 @@
...
+   code[class*="language-"],
+   pre[class*="language-"] {
+     text-shadow: none;
+   }
+ }
+
+ /* Code blocks */
+ pre[class*="language-"] {
+   padding: 1em;
+   margin: .5em 0;
```
Blackbaud-BrandonJones: This seems to be creating a "gap" at the top of the code-block between the header and the block.
screen shot 2018-11-13 at 1 37 52 pm

src/app/public/src/modules/code-block/code-block.component.scss
Blackbaud-BrandonJones: As a side note here, swapping the background color to white, I am curious about your thoughts on making the header color itself the old background color, and adjusting the text color to be accessible still.  

I would love to have @Blackbaud-ToddRoberts weigh in on the color choices for this PR

Blackbaud-BrandonJones: screen shot 2018-11-13 at 1 42 23 pm
I think the current font color and the `$stache-background-color-selected` pass all but AAA on normal text, which I think is passing for us?

Blackbaud-TimPepper: Todd and I discussed making the header a different color, but decided on just making the whole component white. I believe that adding color to the header also introduces more contrast issues with the language label, but I'd have to confirm. 

Adding #fff to the background and header is a good idea though, I'll deliver that now.

src/app/public/src/modules/code-block/prism-a11y.scss
Blackbaud-TimPepper: Good catch, I'll evaluate this now.

src/app/public/src/modules/code-block/code-block.component.scss
Blackbaud-BrandonJones: It passes all but AAA normal text with the language label.  Though if UX believes all white is better than a contrast header that's fine.

Blackbaud-ToddRoberts: I think the white header has enough distinction from the content due to the border, and is low visual weight to distract from the surrounding content. I liken it to a sort of combo between a tile header and list toobar which both have white bgs

PR #556, opened 2018-11-21
update sidebar alignment

Blackbaud-TimPepper:


src/app/public/src/modules/sidebar/sidebar.component.html
```diff
@@ -1,11 +1,11 @@
 <div class="stache-sidebar">
-  <h1 role="navigation" *ngIf="heading" [attr.aria-current]="isHeadingActive() ? 'page' : false" [attr.aria-labelledby]="'stache-sidebar-heading'">
-    <span [ngClass]="{ 'active': isHeadingActive() }" class="stache-sidebar-heading">
+  <span role="navigation" *ngIf="heading" [attr.aria-current]="isHeadingActive() ? 'page' : false" [attr.aria-labelledby]="'stache-sidebar-heading'">
```
Blackbaud-BrandonJones: Is there a reason the span can't be inside the h1?  Also a curious thing I noticed, I am wondering if the `aria-current` should be on the `a` tag instead? Not sure myself, just curious.

Blackbaud-TimPepper: Yes the span being inside the h1 tag was causing about 10px of invisible padding, which was the origin of the alignment issue. Removing that fixed the issue. I can look in to the aria-current attribute

PR #558, opened 2018-11-29
update changelog and version

Blackbaud-TimPepper:
If some writers could give feedback on the changelog text, that would be helpful!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,11 @@
 # 2.14.0 (2018-11-29)
-- Added `stache-hide-from-search` component to allow users to remove content from Stache Search results. [#557](https://github.com/blackbaud/stache2/pull/557)
-- Fixed alignment issue with sidebar and table of contents alignment. [#556](https://github.com/blackbaud/stache2/pull/556)
-- Updated version of `skyux-builder`.
-- Updated Travis build options to enhance test utility.
-- Updated skyux dependencies to depend on components directly. [#553](https://github.com/blackbaud/stache2/pull/553)
-- Updated to allow `Action Button` comopnent to be used without icons.
-- Updated `Action Button` component to be accessible. [#548](https://github.com/blackbaud/stache2/pull/548)
+- Added `Hide From Search` component to allow users to exclude content from Stache search results. [#557](https://github.com/blackbaud/stache2/pull/557)
+- Fixed alignment issue with sidebar and table of contents. [#556](https://github.com/blackbaud/stache2/pull/556)
+- Updated dependencies to use SKY UX components directly. [#553](https://github.com/blackbaud/stache2/pull/553)
+  - Updated version of SKY UX Builder.
+  - Updated Travis build options to enhance test utility.
+- Updated to allow `Action Button` comopnent to be used without icons. [#548](https://github.com/blackbaud/stache2/pull/548)
+  - Updated `Action Button` component to be accessible.
 - Update `Code Block` component to use accessible code highlighting coloring. [#551](https://github.com/blackbaud/stache2/pull/551)
 
 # 2.13.0 (2018-10-23)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,8 @@
 - Updated dependencies to use SKY UX components directly. [#553](https://github.com/blackbaud/stache2/pull/553)
   - Updated version of SKY UX Builder.
   - Updated Travis build options to enhance test utility.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+# 2.14.0 (2018-11-29)
+- Added `stache-hide-from-search` component to allow users to remove content from Stache Search results. [#557](https://github.com/blackbaud/stache2/pull/557)
```
blackbaud-johnly: In the Component docs, we don't format component names as code snippets. It probably doesn't matter here in the changelog, but if "Added hide from search component..." works just as well, it would be more consistent with the formatting in the Component docs.

Not sure if it matters, but should "remove" be "exclude"? I mean, are we excluding the content from the search altogether, or are we including it in the search and then removing it from the search results? I assume the former. If so, "exclude" might be the better word choice here.

Also, "search" at the end should not be capitalized.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
...
+- Fixed alignment issue with sidebar and table of contents alignment. [#556](https://github.com/blackbaud/stache2/pull/556)
```
blackbaud-johnly: We don't need both instances of "alignment." I'd delete the second one, but you can delete the first instead if you think that reads better.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
...
+- Updated version of `skyux-builder`.
```
blackbaud-johnly: Elsewhere in the changelog, we just say "SKY UX Builder" with that capitalization and without the code snippet formatting. I think we should use that style here as well unless there is a reason to format differently in this instance. 

Also, should there be a link to a PR here?

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
...
+- Updated Travis build options to enhance test utility.
```
blackbaud-johnly: Should there be a link to a PR here?

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
...
+- Updated skyux dependencies to depend on components directly. [#553](https://github.com/blackbaud/stache2/pull/553)
```
blackbaud-johnly: "skyux" should be "SKY UX"

FWIW, "dependencies to depend on" seems awkward. And I'm not sure what we mean by it.

CHANGELOG.md
blackbaud-johnly: Actually, disregard the first point about the code snippet. It's not how we format component names in the docs, but it's consistent with the rest of the changelog. We might want to get those on the same page at some point, but we should be consistent within the changelog for now.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
...
+- Updated to allow `Action Button` comopnent to be used without icons.
```
blackbaud-johnly: Should there be a link to a PR here? I assume we should combine this bullet with the next bullet.

The placement of "to allow" is awkward and hard to follow. "Action Button" should not be capitalized; it should be the element name, which I think should be "sky-action-button" here. Also, "component" is misspelled. ... I'd combine this item with the next one as:

"Updated the `sky-action-button` component to improve accessibility and allow action buttons without icons."

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
...
+- Updated `Action Button` component to be accessible. [#548](https://github.com/blackbaud/stache2/pull/548)
```
blackbaud-johnly: See previous comment.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
...
+- Update `Code Block` component to use accessible code highlighting coloring. [#551](https://github.com/blackbaud/stache2/pull/551)
```
blackbaud-johnly: "Updated" should be past tense like the other items. "Code Block" should not be capitalized; it should be the element name, which should be "stache-code-block" here. Also, "accessible code highlighting coloring" is a mouthful and hard to follow. Based on the PR description, I'd change this entry to "Updated `stache-code-block` component to improve accessibility." If you think we need to say more about the specific accessibility change, I'd "Updated `stache-code-block` component to improve accessibility by increasing the color contrast."

CHANGELOG.md
Blackbaud-TimPepper: The ones that don't have a direct link to a PR are all included in the following entries. There's two instances in this update where one PR fixed several things.

Blackbaud-TimPepper: Changed the formatting to clarify this

PR #563, opened 2018-12-12
seperated out clipboard, added comments for code block deprecation

Blackbaud-BrandonJones:


package.json
```diff
@@ -32,16 +32,18 @@
...
     "btoa": "1.1.2",
     "es6-object-assign": "1.1.0",
     "lodash.get": "4.4.2",
     "prismjs": "1.15.0"
   },
   "devDependencies": {
-    "@blackbaud/skyux": "2.30.0",
-    "@blackbaud/skyux-builder": "1.29.0",
-    "@blackbaud/skyux-lib-testing": "1.0.0"
+    "@blackbaud/skyux": "2.33.0",
```
Blackbaud-SteveBrush: We had a hiccup in our release schedule. The latest version is actually `2.33.1`.

src/app/public/src/modules/clipboard/clipboard.component.ts
```diff
@@ -15,33 +17,4 @@ export class StacheCopyToClipboardComponent implements OnInit {
 
   @Input()
   public buttonClickedText: string;
-
-  public buttonActive: boolean = false;
-  public enabled: boolean = false;
```
Blackbaud-SteveBrush: Should you keep `enabled`? Everything that's public on the component is technically a breaking change when you remove it.

src/app/public/src/modules/grid/column.component.ts
```diff
@@ -1,3 +1,7 @@
+/**
+ * Deprecated, update and use the @skyux/fluid-grid unless major bugs are discovered before full deprecation in v3.
```
Blackbaud-SteveBrush: Not a big deal, but could use the JSDoc `@deprecated` annotation.
http://usejsdoc.org/tags-deprecated.html

src/app/public/src/modules/clipboard/clipboard.component.ts
Blackbaud-BrandonJones: It's no longer used on the component and wasn't an input.  It was used for a show hide on the button if the copy option was not available, with the recent update to use the pollyfill it's always available on all devices.

Issue #501, opened 2018-07-09
Wrapping sidebar title padding not respected


Blackbaud-BrandonJones:
When a title item in the Sidebar wraps to new lines, the inline `a` element loses respect for the padding except when applied to the first and last lines of text.

https://docs.blackbaud.com/commerce-onboarding-docs/learn

PR #568, opened 2019-01-17
Table of contents fix

Blackbaud-TimPepper:
Includes Sidebar Heading padding fix for: https://github.com/blackbaud/stache2/issues/501
Fixes several issues with ToC loading:
- Issue where ToC doesn't track user's progress down page if page height updates dynamically
- Issue where ToC will display duplicate links if Angular router caches a component causing ngOnDestroy to not be called


src/app/public/src/modules/sidebar/sidebar.component.scss
```diff
@@ -11,6 +11,7 @@
   border-left: 4px solid transparent;
 
   .stach-sidebar-heading-link {
+    display: block;
```
Blackbaud-TimPepper: Fixes: https://github.com/blackbaud/stache2/issues/501

Blackbaud-BrandonJones:
So after looking and thinking this over, I really believe the offset top has to be the responsibility of the page anchor themselves.  The TOC should not be tied to a specific template decision that's outside of it's scope.  In this case, this change would require the TOC to only be compatible with elements on the page with the `stache-page-anchor` class.  It would also require them to be found when queried in the same specific order as the routes are passed in themselves.  Still up for discussion where/how the trigger would work, but the anchors themselves should maintain and update their own offset.  This way, there can be other uses for the page anchors that may also require the offset in order to perform some functionality.  This would prevent duplication of that effort later.  We can discuss and white board solutions and ideas if you'd like.

Blackbaud-BrandonJones:
The logic for triggering the state of the isCurrent route might also make more sense in the Nav service itself instead of the TOC, since it's the Nav component that's displaying the "active" vs "inactive" state.    the TOC may just be a dumb wrapper of sorts, just a place holder for where that Nav can sit.

src/app/public/src/modules/nav/nav.service.ts
```diff
@@ -46,6 +50,49 @@ export class StacheNavService {
...
+      this.headerHeight = stacheWrapper.offsetTop;
+    }
+  }
+
+  private trackPageOffset() {
+    this.getHeaderHeight();
+
+    // Represents top of page
+    // headerHeight buffer removes discrepency when hero or other 'header' elements exist
+    // 50px buffer aids in click-to-anchor and user visualization
```
Blackbaud-BrandonJones: We can use the `StacheOmnibarAdapterService` to determine if the Omnibar exists.  This service also has the height of the omnibar `50` stored within it.  For components that need to change behavior or height based on the omnibars height, we want to use the value from that service.  This way, in the event the omnibar changes height, we can update the value in one place and all affected components will be updated.

Blackbaud-TimPepper: After looking back in to this logic the code is actually interpreting not only the omnibar height, but also the Stache Hero height, since the Stache Hero is meant to sit outside of the Stache Wrapper.

src/app/public/src/modules/nav/nav.component.ts
```diff
@@ -13,14 +11,11 @@ import { StacheRouteService } from '../shared';
...
   public routes: StacheNavLink[];
-
   @Input()
   public navType: string;
-
   public classname: string = '';
 
-  public constructor(
-    private routerService: StacheRouteService) { }
+  public constructor(private routerService: StacheRouteService) {}
```
Blackbaud-BrandonJones: Know this wasn't you, but noticing it now.  `routerService` is an instance of `StacheRouteService`, we should decide on `route` vs `router`

src/app/public/src/modules/nav/nav.component.ts
```diff
@@ -41,14 +36,11 @@ export class StacheNavComponent implements OnInit, StacheNav {
   private assignActiveStates() {
     const activeUrl = this.routerService.getActiveUrl();
     if (this.hasRoutes()) {
-      this.routes.forEach((route: any) => {
+      this.routes.forEach((route: any, index: number) => {
```
Blackbaud-BrandonJones: index no longer being used?

src/app/public/src/modules/nav/nav.component.ts
```diff
@@ -41,14 +36,11 @@ export class StacheNavComponent implements OnInit, StacheNav {
...
         if (this.isActive(activeUrl, route)) {
           route.isActive = true;
         }
-
-        if (this.isCurrent(activeUrl, route)) {
-          route.isCurrent = true;
-        }
+        route.isCurrent = this.isCurrent(activeUrl, route);
```
Blackbaud-BrandonJones: Much cleaner 👍

src/app/public/src/modules/nav/nav.module.ts
```diff
@@ -12,9 +13,11 @@ import { StacheLinkModule } from '../link';
   imports: [
     CommonModule,
     RouterModule,
-    StacheLinkModule
+    StacheLinkModule,
+    StachePageAnchorModule
```
Blackbaud-BrandonJones: Why are we importing the PageAnchorModule here?

src/app/public/src/modules/nav/nav.service.ts
Blackbaud-BrandonJones: I think the Stache Hero is accounted for with the `headerHeight`, the 50px is from the omnibar, and should use the StacheOmnibarAdapaterService to get that height value.

src/app/public/src/modules/nav/nav.service.ts
```diff
@@ -46,14 +50,59 @@ export class StacheNavService {
...
+
+  // Updates ng-class level trigger to show/hide blue border marking location on page
+  public updateView(routes: StacheNavLink[]) {
+    this.trackPageOffset();
+    this.isCurrent(routes);
+  }
+
+  // Checks to see if stache-wrapper exists and provides offset from top of window
+  private getHeaderHeight() {
+    const stacheWrapper = this.windowRef.nativeWindow.document.querySelector('.stache-wrapper');
```
Blackbaud-BrandonJones: What happens if there is not Stache Wrapper, and they try to use the TOC stand alone?

src/app/public/src/modules/nav/nav.service.ts
```diff
@@ -46,14 +50,59 @@ export class StacheNavService {
...
+    }
+  }
+
+  private trackPageOffset() {
+    this.getHeaderHeight();
+
+    // Represents top of page
+    // headerHeight buffer removes discrepency when hero or other 'header' elements exist
+    // 50px buffer aids in click-to-anchor and user visualization
+    this.pageOffset = ((this.windowRef.nativeWindow.pageYOffset - this.headerHeight) + 50);
```
Blackbaud-BrandonJones: Can we rename this variable to make it more self descriptive and remove the comments?

src/app/public/src/modules/page-anchor/page-anchor.component.ts
```diff
@@ -18,92 +16,29 @@ import { StachePageAnchorService } from './page-anchor.service';
...
-        return element;
-      }
-      element = element.parentElement;
-    }
-    return undefined;
+    const anchor = this.anchorService.generateAnchor(this.elementRef.nativeElement, this.anchorId);
+    this.name = anchor.name;
+    this.fragment = anchor.fragment;
+    this.path = anchor.path as any;
+    this.cdr.detectChanges();
```
Blackbaud-BrandonJones: Change Detector ref still needed here?

src/app/public/src/modules/nav/nav.service.ts
Blackbaud-TimPepper: Currently the ToC isn't set up to be used outside of the context of the wrapper layouts. That's not something we took in to consideration when initially building the component, so I think it's out of scope for what this PR is supposed to be fixing. I think we'd probably want to consider this when/if we decide to break it out. Or at least think about handling this at a later time, otherwise.

src/lib/auth-permissions/index.ts
```diff
@@ -0,0 +1,2873 @@
+/**
```
Blackbaud-BrandonJones: This seems to have been accidentally checked in?

src/app/public/src/modules/page-anchor/page-anchor.component.ts
```diff
@@ -1,109 +1,118 @@
...
+  }
+
+  private getFragment(name: string, anchorId: string): string {
+    return anchorId || name
+      .toLowerCase()
+      .replace(/ /g, '-')
+      .replace(/[^\w-]+/g, '');
+  }
+
+  // TODO: This might not sort properly, must check
```
Blackbaud-BrandonJones: Why are we sorting all the page anchors within the page anchors instead of in the anchor service?  I would suggest either the service, or the subscriber of the `pageAnchors[]` in this case the wrapper, be responsible for sorting?

src/app/public/src/modules/nav/nav-link.ts
```diff
@@ -10,4 +10,5 @@ export interface StacheNavLink {
   isActive?: boolean;
   isCurrent?: boolean;
   showInNav?: boolean;
+  element?: HTMLElement;
```
Blackbaud-BrandonJones: Should no longer be needed

src/app/public/src/modules/footer/footer.component.spec.ts
```diff
@@ -129,7 +129,8 @@ describe('StacheFooterComponent', () => {
     let mappedFooterLinks = footerConfig.nav.items.map((navItem: any) => {
       return {
         name: navItem.title,
-        path: navItem.route
+        path: navItem.route,
+        isCurrent: false
```
Blackbaud-BrandonJones: Why do we need an isCurrent value for footer links?

src/app/public/src/modules/table-of-contents/table-of-contents-wrapper.component.spec.ts
```diff
@@ -100,7 +100,8 @@ describe('Table of Contents Wrapper Component', () => {
         { provide: StacheRouteService, useValue: mockStacheRouteService },
         { provide: SkyAppResourcesService, useValue: mockSkyAppResourcesService },
         { provide: StacheWindowRef, useValue: mockWindowService },
-        { provide: Renderer2, useValue: mockRenderer }
+        { provide: Renderer2, useValue: mockRenderer },
+        StacheOmnibarAdapterService
```
Blackbaud-BrandonJones: We want to use a Mock here so we can avoid cross testing pollution as much as possible.

src/app/public/src/modules/table-of-contents/table-of-contents.component.spec.ts
```diff
@@ -67,40 +92,214 @@ describe('StacheTableOfContentsComponent', () => {
     expect(fixture).toExist();
   });
 
-  it('should detect changes after init', () => {
-    const spy = spyOn((component as any).cdr, 'markForCheck');
-    component.ngAfterViewInit();
-    expect(spy).toHaveBeenCalled();
+  it('should update routes on scroll, undefined routes', () => {
```
Blackbaud-BrandonJones: The statements here should reflect what the test should be doing/expecting so it's obvious when reading test cases.  e.g `should not call updateView when routes are undefined.`

src/app/public/src/modules/table-of-contents/table-of-contents.component.spec.ts
```diff
@@ -67,40 +92,214 @@ describe('StacheTableOfContentsComponent', () => {
...
+    component.updateRoutesOnScroll([]);
+    expect(component.updateView).not.toHaveBeenCalled();
   });
 
-  it('should scroll, route offset <= page offset', () => {
-    mockWindowService.nativeWindow.innerHeight = undefined;
-    component.ngAfterViewInit();
-    component.onScroll();
-    expect(component['activeRoute']).toEqual(route);
+  it('should update routes on scroll, no routes', () => {
```
Blackbaud-BrandonJones: See above

src/app/public/src/modules/table-of-contents/table-of-contents.component.spec.ts
```diff
@@ -67,40 +92,214 @@ describe('StacheTableOfContentsComponent', () => {
...
-      path: '/test2',
-      offsetTop: 200,
-      isCurrent: false
-    });
-    component.onScroll();
-    expect(component['documentBottom']).toEqual(100);
-    expect(component['window'].innerHeight).toEqual(100);
-    expect(component['activeRoute']).toEqual(component.routes[1]);
-    expect(component.routes[1].isCurrent).toBeTruthy();
+  it('should update the view', () => {
```
Blackbaud-BrandonJones: See above

src/app/public/src/modules/table-of-contents/table-of-contents.component.spec.ts
```diff
@@ -67,40 +92,214 @@ describe('StacheTableOfContentsComponent', () => {
...
+    expect(component['documentBottom']).toBe(0);
   });
 
-  it('should scroll, end of page', () => {
-    component.onScroll();
-    expect(component['documentBottom']).toEqual(100);
-    expect(component['window'].innerHeight).toEqual(100);
-    expect(component['activeRoute']).toEqual(route);
-    expect(route.isCurrent).toBeTruthy();
+  it('should update the view, no stache wrapper', () => {
```
Blackbaud-BrandonJones: see above

src/app/public/src/modules/page-anchor/page-anchor.component.ts
```diff
@@ -1,109 +1,106 @@
...
 
-  private getFragment(): string {
-    return this.anchorId || this.name
-      .toLowerCase()
-      .replace(/ /g, '-')
-      .replace(/[^\w-]+/g, '');
-  }
+  public generateAnchor(element: any, anchorId: any = undefined) {
+    this.name = this.getName(element);
+    this.fragment = this.getFragment(this.name, anchorId);
```
Blackbaud-BrandonJones: We should reformat this to.
```
this.fragment = this.anchorId || this.getFragment(this.name);
```
right now there is no need to be passing that anchorId around.

src/lib/auth-permissions/index.ts
Blackbaud-TimPepper: Ah yes this was research I was doing for the show/hide internal docs component. Removing now, good catch.

src/app/public/src/modules/footer/footer.component.spec.ts
Blackbaud-TimPepper: I've been looking in to this and for some reason I cannot immediately identify this the component component is generating a value for this specific test that has the isCurrent and isActive values attached to the generated component.footerLinks. Neither of us has touched this test since I started working on the Stache updates, other than to add those properties to get it running. 

I am going to keep looking in to why this is happening. It feels like a bug.

src/app/public/src/modules/table-of-contents/table-of-contents.component.spec.ts
```diff
@@ -67,40 +92,208 @@ describe('StacheTableOfContentsComponent', () => {
...
-      path: '/test2',
-      offsetTop: 200,
-      isCurrent: false
-    });
-    component.onScroll();
-    expect(component['documentBottom']).toEqual(100);
-    expect(component['window'].innerHeight).toEqual(100);
-    expect(component['activeRoute']).toEqual(component.routes[1]);
-    expect(component.routes[1].isCurrent).toBeTruthy();
+  // it('should update the view when the stache wrapper does not exist', () => {
```
Blackbaud-BrandonJones: Commented out tests

src/app/public/src/modules/wrapper/wrapper.component.ts
```diff
@@ -61,25 +61,35 @@ export class StacheWrapperComponent implements OnInit, OnDestroy, AfterViewInit
...
     this.omnibarService.checkForOmnibar();
     this.jsonData = this.dataService.getAll();
-    this.registerPageAnchors();
+    if (!this.inPageRoutes) {
+      this.pageAnchorService
+        .anchorStream
+        .takeUntil(this.ngUnsubscribe)
+        .subscribe((anchors: BehaviorSubject<StacheNavLink>[]) => {
+        this.inPageRoutes = anchors
+          .map((anchor: BehaviorSubject<StacheNavLink>) => anchor.getValue());
```
Blackbaud-BrandonJones: I think this is the final change,  we need to add a `this.cdr.detectChanges();` after we update the `inPageRoutes`, so the Wrapper component is aware of the change to this field and can pass it down to the children.

src/app/public/src/modules/page-anchor/page-anchor.service.ts
```diff
@@ -1,14 +1,67 @@
...
-import { Subject, Observable } from 'rxjs';
+import { Injectable, OnDestroy } from '@angular/core';
 import { StacheNavLink } from '../nav';
+import { BehaviorSubject, Subject } from 'rxjs';
+import { StacheWindowRef } from '../shared';
 
 @Injectable()
-export class StachePageAnchorService {
+export class StachePageAnchorService implements OnDestroy {
+  public anchorStream$ = new Subject <StacheNavLink[]>();
```
Blackbaud-SteveBrush: What is y'all's preference for denoting an observable property? I've seen either the suffix `Stream` _or_ the use of the dollar sign `$`, but not both? Whichever style you choose, it would probably be good to apply the style to the public property `anchor` (e.g., `anchorStream` or `anchor$`) on the `page-anchor.component.ts` class, too.

Blackbaud-BrandonJones: Is there already a standard pattern used by SKY UX at this time?

Blackbaud-BrandonJones: We've decided to follow the SKY UX pattern of `<property>Stream`.

PR #578, opened 2019-02-11
[HOLD] Initial pass at migrating to SKY UX 3

Blackbaud-BobbyEarl:
Related: https://github.com/blackbaud/skyux-builder-plugin-stache/pull/50


Blackbaud-BrandonJones:
Biggest question, Will this be a major version bump for the Stache library when released as well?  If so, I think there needs to be a meeting to discuss some migration/future Stache plans first.

Blackbaud-BobbyEarl:
Yes it would be a major version release.  This work was done as a test of the migrate plugin.  We did find a few other items that the plugin wasn't accounting for.  I didn't want to lose the work which is why I created this PR.  It is not meant to necessarily dictate the release cycle of Stache.

Having said that, there are known issues with the current version of `@blackbaud/skyux-builder`.  This is due to our dependencies.  These issues have been fixed in later versions of those dependencies, and we reference those newer versions in our next version of builder - `@skyux-sdk/builder`.  That is all to say, I think this a change you will want to consider adopting sooner rather than later.

PR #585, opened 2019-03-27
Update offset and text

Blackbaud-BrandonJones:
When angular loads elements onto the page, such as the Stache images.  The body height doesn't register as changed, the anchors were not updating their offsetTop,  this caused the TOC to be out of sync on pages that included other angular elements as they loaded.

example: https://host.nxt.blackbaud.com/scd-team-docs/style/examples


src/app/public/src/modules/page-anchor/page-anchor.service.ts
```diff
@@ -7,21 +7,20 @@ import { StacheWindowRef } from '../shared';
...
       .map(e => this.windowRef.nativeWindow.document.body.scrollHeight)
       .pairwise()
       .subscribe(height => {
         if (height[0] !== height[1]) {
-          this.bodyHeightChanged.next(height[1]);
+          this.refreshAnchors();
         }
       });
-  }
+   }
```
Blackbaud-SteveBrush: Got a rogue space, here.

src/app/public/src/modules/page-anchor/page-anchor.service.ts
```diff
@@ -7,21 +7,20 @@ import { StacheWindowRef } from '../shared';
 export class StachePageAnchorService implements OnDestroy {
   public anchorStream$ = new Subject <StacheNavLink[]>();
   public pageAnchors: BehaviorSubject<StacheNavLink>[] = [];
-  public bodyHeightChanged = new BehaviorSubject<number>(0);
+  public refreshRequested = new Subject();
```
Blackbaud-SteveBrush: Changing public methods and properties on a service is technically a breaking change. Would it still make sense to retain the name `bodyHeightChanged`?

Blackbaud-BrandonJones: This is a change to the unreleased branch so it will not be a breaking change.

src/app/public/src/modules/page-anchor/page-anchor.service.ts
Blackbaud-BrandonJones: good catch!

PR #586, opened 2019-04-03
Stache internal view

Blackbaud-TimPepper:
Create a way for users to show/hide content based on the availability of a BBID. 

Both directive and component options are shown here, though we can decide on one or the other or go with both.


src/app/public/src/modules/internal/auth.service.ts
```diff
@@ -0,0 +1,26 @@
...
+@Injectable()
+export class StacheAuthService  {
+
+  public isAuthenticated = new BehaviorSubject<boolean>(false);
+
+  constructor(private auth: SkyAuthTokenProvider) {
+    this.checkForAuth();
+  }
+
+  public checkForAuth() {
```
Blackbaud-BobbyEarl: Any reason for this to be public?

PR #590, opened 2019-04-11
update sidebar component to support updating dynamically

Blackbaud-TimPepper:


src/app/public/src/modules/sidebar/sidebar.component.ts
```diff
@@ -8,38 +8,39 @@ import { StacheRouteService } from '../shared';
...
+export class StacheSidebarComponent implements StacheNav {
   @Input()
-  public routes: StacheNavLink[];
+  public set routes(value: StacheNavLink[]) {
+    this._routes = value ? value : this.routeService.getActiveRoutes();
+    this.childRoutes = this.filterRoutes(this._routes);
+  }
+
+  public get routes(): StacheNavLink[] {
+    return this._routes || [];
```
Blackbaud-BrandonJones: Current tests never hit this "else" case.

src/app/public/src/modules/sidebar/sidebar.component.ts
```diff
@@ -8,38 +8,39 @@ import { StacheRouteService } from '../shared';
...
     const url = this.routeService.getActiveUrl();
     return (url === this.headingRoute);
   }
 
   private filterRoutes(routes: StacheNavLink[]): StacheNavLink[] {
-    const root = routes[0];
-    let headingPath = Array.isArray(root.path) ? root.path.join('/') : root.path;
-    headingPath = headingPath.replace(/^\//, '');
-    this.heading = root.name;
+    if (routes) {
```
Blackbaud-BrandonJones: current tests don't hit this "else" case

src/app/public/src/modules/sidebar/sidebar.component.ts
Blackbaud-BrandonJones: I think with the current `set` rules, It will never return the [] assignment in the get.  Likely unnecessary.

src/app/public/src/modules/sidebar/sidebar.component.ts
Blackbaud-BrandonJones: I think the routerService will always return a base route for the current path.  So this if check might not be necessary.

src/app/public/src/modules/sidebar/sidebar.component.spec.ts
```diff
@@ -57,18 +57,16 @@ describe('StacheSidebarComponent', () => {
...
       declarations: [
-        StacheNavComponent,
-        StacheSidebarComponent,
+        SidebarTestComponent,
         RouterLinkStubDirective
       ],
       imports: [
         RouterTestingModule,
-        StacheLinkModule
+        StacheModule
```
Blackbaud-BrandonJones: We can be more isolated here and just import the sidebar module.

src/app/public/src/modules/sidebar/sidebar.component.spec.ts
```diff
@@ -147,20 +141,36 @@ describe('StacheSidebarComponent', () => {
...
+  });
+
+  it('should use generated routes if no custom routes are provided', () => {
+    const tmpFixture = TestBed.createComponent(SidebarTestComponent);
+    const el = tmpFixture.nativeElement as HTMLElement;
+
+    tmpFixture.detectChanges();
+
+    const sidebarHeading = el.querySelector('#stache-sidebar-heading').innerHTML;
+    console.log(sidebarHeading);
```
Blackbaud-BrandonJones: console.log

src/app/public/src/modules/sidebar/sidebar.component.spec.ts
```diff
@@ -147,20 +141,36 @@ describe('StacheSidebarComponent', () => {
...
     expect(component.routes[0].name).toBe('Home');
   });
+
+  it('should use generated routes if no custom routes are provided', () => {
+    spyOn(component['routeService'], 'getActiveRoutes').and.callThrough();
+    component.routes = undefined;
+    expect(component['routeService'].getActiveRoutes).toHaveBeenCalled();
+  });
+
+  it('should use generated routes if no custom routes are provided', () => {
```
Blackbaud-BrandonJones: These two test case descriptors are the same?

src/app/public/src/modules/sidebar/sidebar.component.spec.ts
```diff
@@ -147,20 +142,35 @@ describe('StacheSidebarComponent', () => {
...
     expect(component.routes[0].name).toBe('Home');
   });
+
+  it('should use generated routes if no custom routes are provided', () => {
+    spyOn(component['routeService'], 'getActiveRoutes').and.callThrough();
+    component.routes = undefined;
+    expect(component['routeService'].getActiveRoutes).toHaveBeenCalled();
+  });
+
+  it('should use generated routes if custom routes are provided', () => {
```
Blackbaud-BrandonJones: Should this be "should `not` use generated routes..." ?

src/app/public/src/modules/sidebar/sidebar.component.spec.ts
```diff
@@ -147,20 +142,45 @@ describe('StacheSidebarComponent', () => {
...
   it('should add a heading for the root level route', () => {
-    fixture.detectChanges();
+    component.routes = undefined;
     expect(component.heading).toEqual('Home');
     expect(component.headingRoute).toEqual('/');
   });
 
   it('should automatically add a link to the top-level page', () => {
-    fixture.detectChanges();
+    component.routes = undefined;
```
Blackbaud-BrandonJones: Should no longer be necessary to set these to undefined.  May have to detectChanges instead.

PR #593, opened 2019-05-02
Feature lots for migrate

Blackbaud-BrandonJones:
Current progress of migration, key areas to review are `package.json` dependencies, imports/exports of services that are available.


package.json
```diff
@@ -22,28 +22,57 @@
...
-    "@blackbaud/skyux-lib-code-block": "1.0.0-rc.3",
-    "@blackbaud/skyux-builder-plugin-stache": "~1.8.0",
-    "@blackbaud/skyux-builder-stache-search": "^1.7.0",
-    "btoa": "1.1.2",
     "es6-object-assign": "1.1.0",
     "lodash.get": "4.4.2",
-    "prismjs": "1.15.0"
+    "prismjs": "1.15.0",
+    "btoa": "1.1.2",
+    "zone.js": "0.8.28"
```
Blackbaud-SteveBrush: Should only be listed in `devDependencies`.

package.json
```diff
@@ -22,28 +22,57 @@
...
   "peerDependencies": {
-    "@angular/common": "^4.0.0",
-    "@angular/compiler": "^4.0.0",
-    "@angular/core": "^4.0.0",
-    "@angular/forms": "^4.0.0",
-    "@angular/platform-browser": "^4.0.0",
-    "@angular/platform-browser-dynamic": "^4.0.0",
-    "@angular/router": "^4.0.0",
-    "@blackbaud/skyux": "^2.0.0"
+    "@angular/common": ">=4.0.0",
```
Blackbaud-SteveBrush: Is Stache importing from all of these packages directly (for all files within the `./src/app/public` directory)? If so, I would list the version numbers at the lowest possible supported version, so that the majority of SKY UX 3 SPAs can consume Stache without incurring a "missing peer" warning during `npm install`.

For example, does Stache work with `@skyux/core@3.0.0`? If so, you would list it as `^3.0.0`.

